PR #10, opened 2026-03-21
fix(panel): resolve Electron dev mode startup freeze and blank window

atreidesfamilyatomics:
## Summary

The Electron panel (`npm run dev`) had a combination of issues that made the app
unresponsive or blank on first launch in development mode. This PR fixes four
independent root causes, all isolated to `panel/src/main/` and the renderer.

---

### 1. Main thread blocked before window creation (`engine-manager.ts`)

`checkEngineVersion()` called `hashSourceFiles()` synchronously — SHA256-hashing
every Python source file in the engine — before `createWindow()`. In development
mode this work is meaningless (the developer manages their own engine installation),
but it blocked the Electron main thread long enough to delay IPC readiness and produce
a visually frozen window.

**Fix:** Short-circuit `checkEngineVersion()` when `!app.isPackaged`, returning
`{ needsUpdate: false }` immediately. Also added 5-second timeouts to all
`getVersionFromBinary()` calls and replaced a shell `exec()` fallback with safe
`execFile` + promisify to eliminate shell injection risk.

### 2. Blocking `ps aux` scan after window creation (`sessions.ts`)

`detectAndAdoptAll()` called `execSync('ps aux')` in the main process — a blocking
scan that could stall the main thread for 1–5 s on a loaded system — immediately
after the window was created but before the renderer had finished mounting.

**Fix:** Replaced `execSync` with `execFile` + `promisify`, making the scan async
with a 5-second timeout. The `detect()` method was already declared `async` so no
callers needed updating.

### 3. Blank white window in dev mode (`index.ts`)

Vite's `@vitejs/plugin-react` injects an inline script preamble required for React
Fast Refresh. The existing Content Security Policy (`script-src 'self'`) blocked this
inline script, preventing React from mounting and producing a completely blank window.

**Fix:** Wrapped the entire CSP `onHeadersReceived` handler in `if (app.isPackaged)`.
The CSP is meaningless in dev mode (assets are served by Vite's dev server, not
Electron's file protocol) and was always the wrong place to apply it for development.

### 4. Renderer hangs if IPC response is slow (`App.tsx`)

The startup `checkInstallation()` IPC call had no timeout. If the main process was
temporarily busy, the renderer would block indefinitely on `checkingSetup === true`
and render nothing. The original guard also returned `null` while checking, producing
a blank screen with no feedback.

**Fix:** Wrapped the IPC call in a `Promise.race` with a 6-second timeout so
`.finally()` always fires and `checkingSetup` always clears. Replaced `return null`
with a minimal CSS spinner so users see the app is alive during the check.

### 5. `additionalArgs` comma-list causes argparse failure (`sessions.ts`)

When users copy flag examples from documentation (e.g. `--flag1, --flag2`), the
comma-separated format produces malformed tokens after the whitespace split:
`['--flag1,', '--flag2']`. A bare flag name at the end of the args array causes
Python argparse to error with `expected one argument`.

**Fix:** Strip commas from `additionalArgs` before splitting on whitespace.

---

## Files Changed

| File | Change |
|------|--------|
| `panel/src/main/engine-manager.ts` | Dev mode short-circuit; async timeouts in binary version check |
| `panel/src/main/sessions.ts` | Async `ps aux`; comma-stripping in `additionalArgs` |
| `panel/src/main/index.ts` | CSP guard: `if (app.isPackaged)` |
| `panel/src/renderer/src/App.tsx` | IPC timeout + spinner during installation check |

---

## Testing

- `npm run dev` from `panel/` now opens a responsive window within 2 seconds on macOS
- Window reacts immediately to mouse and keyboard input
- Model loads, serves, and tool calls function correctly
- Packaged builds (`npm run dist`) are unaffected — all fixes use `app.isPackaged` guards
  or are unconditionally safe

---

> **Note:** This branch also includes the QuantizedKVCache list/tuple fix from
> [#9](https://github.com/jjang-ai/vmlx/pull/9) for convenience, but that change
> is scoped entirely to `vmlx_engine/` and can be reviewed independently.

### DIFF
--- a/panel/src/main/sessions.ts
+++ b/panel/src/main/sessions.ts
@@ -1803,7 +1803,8 @@ export class SessionManager extends EventEmitter {
     // Additional arguments — strip stale image-only flags from old session configs
     if (config.additionalArgs?.trim()) {
       const staleImageFlags = new Set(['--mflux-class', '--image-mode', '--image-quantize'])
-      const extra = config.additionalArgs.trim().split(/\s+/).filter(Boolean)
+      // Strip commas — common when users copy flag lists from docs (e.g. "--flag1, --flag2")
+      const extra = config.additionalArgs.trim().replace(/,/g, ' ').split(/\s+/).filter(Boolean)
       const filtered: string[] = []
       for (let i = 0; i < extra.length; i++) {
         if (staleImageFlags.has(extra[i])) {


jjang-ai:
Thanks for the clean PR. Four independent, well-scoped fixes — each one makes sense.

Quick review:

1. **`checkEngineVersion` short-circuit in dev mode** — Good. No reason to hash source files when `!app.isPackaged`. The 5s timeout on `getVersionFromBinary` is a nice safety net.

2. **`execSync('ps aux')` → async** — Good fix. Blocking the main thread with `execSync` is bad practice. The method was already `async` so this is a clean drop-in.

3. **CSP bypass in dev mode** — Correct. Vite's React Fast Refresh requires inline scripts. CSP in dev mode was always wrong.

4. **IPC timeout + loading spinner** — Good UX improvement. Showing a spinner instead of a blank white screen is much better.

These are all dev-mode only changes that don't affect production builds. Will merge.

PR #16, opened 2026-03-22
fix: hybrid model block cache — Metal panic, reconstruction, and re-prefill (non-MLLM path)

Fail-Safe:
## Summary

> **Note:** These changes are offered as suggestions based on our experience running Qwen3.5-35B-A3B-4bit on Apple Silicon. We understand the codebase is evolving quickly and these fixes may not align with the intended architecture — happy to adjust, split, or withdraw anything here based on your feedback.

**Rebased onto latest `main` (e341eb9).** Single squashed commit for clean review.

This PR addresses issues we encountered with hybrid SSM models (KV + ArraysCache layers) running through the **non-MLLM scheduler path** (`mllm=False`). We understand the MLLM path has `HybridSSMStateCache` for companion SSM state — these fixes address the same gap in the regular `scheduler.py` / `prefix_cache.py` path.

### What we found

On v1.3.7 `main` with `mlx-community/Qwen3.5-35B-A3B-4bit` (which loads as `mllm=False`):
- **Kernel panic** on the second request — Metal `completeMemory() prepare count underflow`
- **~50s post-generation re-prefill** on every cache-hit request (full prompt re-processed just for cache storage)
- **Block cache reconstruction fails** on hybrid models (10 layers reconstructed, 40 expected)
- **Block KV data lost** after cache-hit requests (`skip=40` instead of `kv=10`)

### Fixes (all in `prefix_cache.py` and `scheduler.py`)

1. **Metal kernel panic**: Block cache stored lazy MLX slices of already-evaluated arrays. Three root causes:
   - Numpy pre-conversion was gated on `disk_store is not None`
   - `hasattr(keys, '__array__')` is always `False` for MLX arrays (they use `__dlpack__`) — the numpy safety path was dead code
   - `bfloat16` dtype unsupported by numpy — cast to float16 intermediary, restore original dtype after

2. **Block cache reconstruction miscount**: The miss diagnostic used flat list length to check layer indices, miscounting KV layers at high positions. Fix: track indices in a set.

3. **~50s post-generation re-prefill**: `_truncate_cache_to_prompt_length` returned `None` for hybrid models. Fix: pass SSM layers through unchanged, skip L2 disk writes for hybrid models.

4. **Cache position offset**: After a block cache hit, the BatchGenerator's raw cache only contains newly-processed tokens. Block extraction used global positions, causing all KV slices to be skipped. Fix: detect and apply a position offset.

## Performance comparison

Tested on `mlx-community/Qwen3.5-35B-A3B-4bit`, M3 Max 48GB, rebased onto latest `main` (e341eb9):

**v1.3.7 main (unpatched):** crashes on second request (Metal kernel panic)

**With this PR:**
```
  Turn       TTFT    Total   Prompt   Cached  Speedup
  —————— ———————— ———————— ———————— ———————— ————————
  1         2.93s    5.70s     4244        0        —
  2         215ms    1.77s     4259     4224    13.6x
  3         217ms    3.00s     4287     4224    13.5x
  4         303ms    3.06s     4306     4224     9.7x
```

## Trade-offs and open questions

- **L2 disk cache writes are skipped for hybrid models** since SSM state can't be cleanly truncated to prompt-only. Hybrid models rely on in-memory block cache (L1) only. If there's a preferred approach for persisting hybrid cache state to disk, happy to implement differently.
- **bfloat16 → float16 → bfloat16 round-trip** in the numpy intermediary loses some precision. An alternative might be `mx.contiguous()` to force materialization without numpy, if that avoids the same Metal issues.
- **`hasattr(keys, '__array__')` being dead code** may affect non-hybrid models too — the numpy safety path was never executing for any MLX array. Worth checking if this was intentional or a latent bug.
- **Block disk store `std::bad_cast`**: We see `Pre-serialize/write failed for block: std::bad_cast` on upstream `main` (pre-existing, not caused by this PR). The in-memory block cache works correctly regardless.

## Test results

- 207/207 cache-related tests pass
- 2089/2111 total tests pass (22 failures are pre-existing on upstream `main`)
- No kernel panics across multiple conversation sessions
- No "Re-prefilled N tokens" log messages

### DIFF
--- a/vmlx_engine/prefix_cache.py
+++ b/vmlx_engine/prefix_cache.py
@@ -362,6 +362,78 @@ class PrefixCacheManager:
 # =============================================================================
 
 
+
+def _numpy_block_slice(
+    cache_data, np_sources, start_idx, end_idx, is_last_block,
+):
+    """Create per-block cache_data using numpy slicing (no MLX/Metal ops).
+
+    Args:
+        cache_data: Full layer-state dicts from _extract_cache_states.
+        np_sources: dict mapping layer_idx → (np_keys, np_values).
+        start_idx, end_idx: Token range for this block.
+        is_last_block: Whether this is the last block in the sequence.
+
+    Returns:
+        List of tuples per layer in the same format as _extract_block_tensor_slice
+        but with numpy arrays instead of MLX arrays.
+    """
+    import numpy as np
+
+    block_slices = []
+    for idx, layer_state in enumerate(cache_data):
+        if "state" not in layer_state:
+            continue
+        cls = layer_state.get("class_name", "")
+
+        # CacheList (MoE): not yet supported in numpy path
+        if cls == "CacheList":
+            block_slices.append(("skip",))
+            continue
+
+        if idx in np_sources:
+            np_k, np_v, *_ = np_sources[idx]
+            ndim = np_k.ndim
+            if ndim == 4:
+                seq_len = np_k.shape[2]
+            elif ndim == 3:
+                seq_len = np_k.shape[1]
+            else:
+                block_slices.append(("skip",))
+                continue
+            actual_end = min(end_idx, seq_len)
+            if start_idx >= actual_end:
+                block_slices.append(("skip",))
+                continue
+            if ndim == 4:
+                ks = np_k[:, :, start_idx:actual_end, :]
+                vs = np_v[:, :, start_idx:actual_end, :]
+            else:
+                ks = np_k[:, start_idx:actual_end, :]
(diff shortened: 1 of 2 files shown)


jjang-ai:
Cherry-picked onto main as f0f9858. Thanks for the excellent contribution — the numpy pre-conversion for Metal safety, position offset fix, and reconstruction index tracking are all solid fixes. The 13.6x TTFT improvement on hybrid models is a great result.

Integrated alongside our v1.3.8 SSD streaming fixes and hybrid SSM safety checks. 2122 tests pass, no regressions.

PR #17, opened 2026-03-22
fix: block disk store bfloat16/numpy serialization for hybrid models

Fail-Safe:
## Summary

> **Note:** Follow-up to #16 (merged). Offered as a suggestion — happy to adjust based on your preferred approach.

`BlockDiskStore.write_block_async()` fails with `std::bad_cast` on every block write for bfloat16 hybrid models (e.g. Qwen3.5-35B-A3B). Two issues:

1. **numpy ndarrays in tensors dict**: The numpy-safe slicing path (from #16) produces numpy arrays in block `cache_data`. `mx.save_safetensors()` can't serialize numpy arrays.

2. **bfloat16 MLX arrays**: The safetensors format doesn't support bfloat16, causing `std::bad_cast` in the C++ layer when blocks reach the disk write path via `paged_cache.py` eviction.

### Fix

In `write_block_async()`, normalize all tensors before saving:
- Convert numpy `ndarray` → `mx.array`
- Cast `bfloat16` → `float16`

On deserialization, restore `float16` → `bfloat16` for KV and rotating_kv layers.

### Impact

Enables cross-session block disk cache persistence for bfloat16 hybrid models. Previously, every block write failed silently with `std::bad_cast` (hundreds of debug messages per request on a 60K+ token conversation).

### Trade-offs

- **float16 round-trip**: bfloat16 → float16 → bfloat16 loses some mantissa precision. For KV cache data this is negligible, but worth noting.
- **Assumes bfloat16 on load**: The deserialization unconditionally casts float16 back to bfloat16 for KV layers. This is correct for bfloat16 models but would be wrong for models that natively use float16. A metadata flag indicating the original dtype would be more robust.

## Test plan

- [x] 82/82 cache-related tests pass
- [ ] Verify blocks persist to disk (no `std::bad_cast` messages)
- [ ] Verify cross-session cache hit after server restart

### DIFF
--- a/tests/test_block_disk_bfloat16.py
+++ b/tests/test_block_disk_bfloat16.py
@@ -1,0 +1,154 @@
+"""Tests for block disk store bfloat16/numpy serialization.
+
+Verifies that bfloat16 KV cache data and numpy arrays survive the
+serialize -> disk -> deserialize round-trip correctly.
+"""
+import pytest
+import tempfile
+import shutil
+from pathlib import Path
+
+try:
+    import mlx.core as mx
+    import numpy as np
+    HAS_MLX = True
+except ImportError:
+    HAS_MLX = False
+
+pytestmark = pytest.mark.skipif(not HAS_MLX, reason="MLX not available")
+
+
+@pytest.fixture
+def disk_store():
+    """Create a temporary BlockDiskStore for testing."""
+    from vmlx_engine.block_disk_store import BlockDiskStore
+    tmpdir = tempfile.mkdtemp(prefix="test_block_disk_")
+    store = BlockDiskStore(tmpdir, max_size_gb=1.0)
+    yield store
+    store.shutdown()
+    shutil.rmtree(tmpdir, ignore_errors=True)
+
+
+def _make_kv_block(seq_len=64, n_heads=2, head_dim=256, dtype=mx.bfloat16):
+    """Create a simple KV-only block cache_data (10 KV layers, 30 skip)."""
+    cache_data = []
+    for i in range(40):
+        if i % 4 == 3:  # 10 KV layers at positions 3,7,11,...
+            keys = mx.random.normal((1, n_heads, seq_len, head_dim)).astype(dtype)
+            values = mx.random.normal((1, n_heads, seq_len, head_dim)).astype(dtype)
+            mx.eval(keys, values)  # noqa: S307 -- mlx tensor materialization
+            cache_data.append(("kv", keys, values))
+        else:
+            cache_data.append(("skip",))
+    return cache_data
+
+
+def _make_numpy_kv_block(seq_len=64, n_heads=2, head_dim=256):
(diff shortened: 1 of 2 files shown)


Fail-Safe:
> Thanks for the thorough work — these PRs are consistently high quality.

@jjang-ai You're very welcome--and thank you for being willing to consider my contributions! In full fairness, I'm leaning heavily on Claude's ability to navigate the intricacies of the MLX framework. I just happen to have a decent background in Python (among other languages) and a relentless desire to keep iterating for improvement's sake. 😎

For future us, an assessment from Claude based on quantitative testing:

```plaintext
======================================================================
bfloat16 -> float16 -> bfloat16 Precision Loss Analysis
(for block disk cache serialization)
======================================================================

1. Random normal distribution (mean=0, std=1)
Simulates typical KV cache attention values
Elements:       32,768
Max abs error:  0.00e+00
Mean abs error: 0.00e+00
Max rel error:  0.000000 (0.0000%)
Mean rel error: 0.000000 (0.0000%)
Values changed: 0.0%
Overflow (>65504): 0

2. Wider distribution (mean=0, std=10)
Stress test for larger attention values
Elements:       32,768
Max abs error:  0.00e+00
Mean abs error: 0.00e+00
Max rel error:  0.000000 (0.0000%)
Mean rel error: 0.000000 (0.0000%)
Values changed: 0.0%
Overflow (>65504): 0

3. Full-scale KV block (1, 2, 64, 256) x 10 layers
Matches Qwen3.5-35B-A3B KV cache dimensions
Total elements: 655,360
Max abs error:  2.98e-08
Mean abs error: 4.55e-14
Max rel error:  0.004695 (0.4695%)
Mean rel error: 0.000000 (0.0000%)
Values changed: 0.0%
Overflow (>65504): 0

======================================================================
Analysis
======================================================================

bfloat16: 1 sign + 8 exponent + 7 mantissa bits (same range as float32)
float16:  1 sign + 5 exponent + 10 mantissa bits (range +/-65504)

Round-trip path: bfloat16 -> float16 -> bfloat16

The round-trip is LOSSLESS for values in the float16 range because:
- bfloat16 has 7 mantissa bits, float16 has 10
- All 7 mantissa bits of bfloat16 are preserved in float16's 10 bits
- Converting back truncates the extra 3 bits (which were zero)

The ONLY loss occurs for values outside float16's range:
- |value| > 65504: overflows to +/-inf in float16
- Very small subnormals: may underflow

For KV cache data (typically |values| < 10), this is lossless.

⏺ The round-trip is effectively lossless. Zero values changed across 655K+ elements. The theoretical explanation checks out:

  - bfloat16 has 7 mantissa bits, float16 has 10 mantissa bits
  - All 7 bfloat16 mantissa bits fit perfectly within float16's 10 bits
  - Converting back truncates only the 3 extra zero-padded bits
  - The only risk is overflow for |values| > 65504, which doesn't occur in KV cache data
```

PR #28, opened 2026-03-25
fix: resolve HuggingFace repo IDs to local paths for MLLM and JANG detection

EliasOenal:
## Summary
When the model is specified as a HuggingFace repo ID (e.g. `JANGQ-AI/Qwen3.5-122B-A10B-JANG_3L`), several file-based detection paths silently fail because they treat the repo ID as a literal filesystem path. This means `is_mllm_model()`, JANG model loading, and tokenizer fallback detection all miss models that are only present in the local HF cache.
## Problem
`is_mllm_model()` checks for `config.json` and `jang_config.json` by joining the model name with `os.path.join()`. When the model name is `Org/Model`, that path doesn't exist on disk — the actual files live in `~/.cache/huggingface/hub/models--Org--Model/snapshots/{hash}/`. The same issue affects `MLXMultimodalLM.load()` (JANG VLM loading) and `load_model_with_fallback()` (JANG text model loading).
## Fix
Add `resolve_to_local_path()` in `vmlx_engine/api/utils.py` that scans the local HF cache via `huggingface_hub.scan_cache_dir()` (no network, no downloads) and returns the snapshot path of the most recently modified revision. Falls back to the original name if resolution fails.
Applied in three places:
- `is_mllm_model()` — vision config detection
- `MLXMultimodalLM.load()` — JANG VLM loading
- `load_model_with_fallback()` — JANG text model loading and tokenizer fallback detection
Results are cached via `@lru_cache` for the lifetime of the process. The model config registry lookup tries the resolved path first (so it can read `config.json`), then falls back to the original name for name-based regex disambiguation (e.g. GLM-Z1, MedGemma patterns).
## Tests
6 new tests in `test_api_utils.py` covering: existing directories returned as-is, unresolvable names returned as-is, HF repo ID resolution from cache, most-recent-revision selection, scan failure graceful fallback, and LRU cache hit verification.

### DIFF
--- a/tests/test_api_utils.py
+++ b/tests/test_api_utils.py
@@ -451,14 +451,13 @@ class TestResolveToLocalPath:
         """HuggingFace repo IDs are resolved to snapshot paths via cache scan."""
         from unittest.mock import MagicMock
         from types import SimpleNamespace
-        from datetime import datetime
 
         snapshot_dir = tmp_path / "snapshots" / "abc123"
         snapshot_dir.mkdir(parents=True)
         (snapshot_dir / "config.json").write_text("{}")
 
         revision = SimpleNamespace(
-            last_modified=datetime(2025, 1, 1),
+            last_modified=1704067200.0,  # 2024-01-01
             snapshot_path=snapshot_dir,
         )
         repo = SimpleNamespace(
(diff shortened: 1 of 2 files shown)


jjang-ai:
Integrated in commit d5db9d2 with 3 fixes (removed return False regression, added server.py JANG metadata call site, resolved path in tokenizer fallback). Thank you @EliasOenal!

PR #29, opened 2026-03-25
fix: add image placeholder tokens for assistant role messages in MLLM

EliasOenal:
## Summary
Fix an error when images are attached to assistant-role messages in multimodal conversations. The vision encoder produces image features but no matching placeholder tokens exist in the prompt, causing a `ValueError` during generation.
## Problem
`_extract_multimodal_messages()` only builds the multimodal content list (with `{"type": "image"}` placeholder entries) for `user`-role messages. When a UI sends images on `assistant` messages — e.g. [klite](https://github.com/LostRuins/lite.koboldai.net) attaches images to conversation context on the assistant turn — the vision encoder produces features for those images but the rendered prompt has no corresponding image tokens:

> **ValueError: Image features and image tokens do not match: tokens: 0, features 900**

## Fix
Extend the image-placeholder branch in `_extract_multimodal_messages()` (`mllm.py:950`) from `role == "user"` to `role in ("user", "assistant")`, so both roles get the `{"type": "image"}` markers that the vision encoder expects. When the assistant message also carries `tool_calls`, those are preserved on the output dict.
## Tests
3 new tests in `test_mllm.py`:
- `test_assistant_image_tokens_included` — assistant message with image produces placeholder and extracts the URL
- `test_assistant_image_with_tool_calls_preserved` — assistant message with both images and tool_calls retains both
- `test_user_image_still_works` — regression guard confirming user-role images are unaffected

### DIFF
--- a/tests/test_mllm.py
+++ b/tests/test_mllm.py
@@ -345,6 +345,82 @@ class TestExtractMultimodalMessages:
         assert "tool_calls" in assistant_msg
         assert assistant_msg["tool_calls"] == tool_calls
 
+    def test_assistant_image_tokens_included(self):
+        """Assistant messages with images must produce image placeholders.
+
+        Regression test: previously only user role got image content lists.
+        Some UIs (e.g. klite) send images on assistant messages when
+        attaching context images to the conversation.
+        """
+        from vmlx_engine.models.mllm import MLXMultimodalLM
+
+        messages = [
+            {
+                "role": "assistant",
+                "content": [
+                    {"type": "text", "text": "Here is the image analysis."},
+                    {"type": "image_url", "image_url": {"url": "https://example.com/img.jpg"}},
+                ],
+            }
+        ]
+
+        chat_msgs, images, _videos = MLXMultimodalLM._extract_multimodal_messages(messages)
+        assert len(images) == 1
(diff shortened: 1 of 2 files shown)


EliasOenal:
@jjang-ai thank you for integrating the fix and for taking the time to polish it up. I wanted to respond to your review, but I hadn't found the time yet. I really appreciate the project: it combines the speed of MLX with a lot of the capabilities people expect from modern inference engines.

On attribution: Git already supports separate `Author` and `Committer` metadata, which is probably the most standard way to preserve who originally wrote a change versus who applied or edited it. `Co-authored-by:` trailers are another common option.

This mostly only matters if you want especially detailed attribution later on. For example for downstream packaging work such as Debian/Ubuntu, where tracing authorship is a requirement. I ran into that on one of my own projects, where it later was a bit of work to reconstruct this, together with the package maintainer. That said, the way you referenced the PR in the integrating commit is overall fine from my perspective.

And one last thing that might be useful to your agent while developing vmlx: I built [term-cli](https://github.com/EliasOenal/term-cli), a terminal for coding agents that handles blocking and interactive workflows like `mlx serve` or `ssh`. My Claude uses it heavily for closed-loop debugging and smoke-testing server applications.

PR #55, opened 2026-04-07
fix: rename switch_mlp to experts.switch_glu for Gemma 4 JANG MoE models

zheli:
## Summary

- Gemma 4 JANG models (26B/31B) produce completely **garbled output** because all MoE expert weights are silently dropped at load time — leaving every expert layer at random initialisation.
- Root cause: JANG safetensors use `switch_mlp.{gate,up,down}_proj.*` key names, but mlx-vlm's `Gemma4TextModel` wraps the experts in `Experts → SwitchGLU` whose parameter paths are `experts.switch_glu.{gate,up,down}_proj.*`. The mismatch causes `load_weights(strict=False)` to discard all expert weights silently.
- Fix: rename any `switch_mlp` key to `experts.switch_glu` in the shard weight dict before calling `load_weights()`.

## Root Cause Detail

`load_weights()` is called with `strict=False` (line 819), so mismatched keys are silently ignored. The naming gap was already noticed — the quantised-suffix tracking block at ~line 554 contains this comment:

```python
# JANG uses switch_mlp.* but model expects experts.switch_glu.*)
if ".switch_mlp." in qpath:
    quantized_suffixes.add(
        qpath.replace(".switch_mlp.", ".experts.switch_glu.")
    )
```

However that block only adds paths to the quantisation set; the rename was never applied to the actual weight dict before loading.

## Changes

**`vmlx_engine/utils/jang_loader.py`**  
Added a rename pass immediately before `model.load_weights()` (inside the per-shard loop):

```python
# Gemma 4 MoE: JANG stores expert weights under switch_mlp.{gate,up,down}_proj
# but mlx-vlm's Gemma4TextModel wraps them in an Experts(nn.Module) whose
# SwitchGLU sub-module expects experts.switch_glu.{gate,up,down}_proj.
# Without this rename the weights don't match any model parameter path and
# are silently discarded by load_weights(strict=False), leaving all MoE
# expert layers at their random initialisation — producing garbled output.
if any(".switch_mlp." in k for k in shard_weights):
    shard_weights = {
        (k.replace(".switch_mlp.", ".experts.switch_glu.") if ".switch_mlp." in k else k): v
        for k, v in shard_weights.items()
    }
```

The rename is gated on the presence of a `switch_mlp` key so it is a no-op for dense models, Nemotron, and any other architecture that does not use this naming convention.

**`tests/test_jang_loader.py`**  
Added `TestGemma4SwitchMlpRename` (6 tests):
- `test_switch_mlp_keys_are_renamed` — no `switch_mlp` survives
- `test_experts_switch_glu_keys_present_after_rename` — all three projections appear at the correct path
- `test_non_switch_mlp_keys_are_unchanged` — dense MLP and router keys are untouched
- `test_key_count_preserved` — 1-to-1 rename, no keys added or lost
- `test_no_rename_when_no_switch_mlp_keys` — no-op for non-MoE shards
- `test_rename_applied_in_jang_loader_source` — guards against future regressions

## Testing

```
pytest tests/test_jang_loader.py -v
# 24 passed in 0.63s
```

Verified end-to-end on `dealignai/Gemma-4-26B-A4B-JANG_4M-CRACK`: the model now produces coherent output instead of garbled text.

### DIFF
--- a/tests/test_jang_loader.py
+++ b/tests/test_jang_loader.py
@@ -1,6 +1,114 @@
 """Tests for JANG model loading — param count, tree_flatten, detection, MCP config."""
+
 import pytest
 from pathlib import Path
+
+
+class TestGemma4SwitchMlpRename:
+    """Verify that Gemma 4 JANG switch_mlp weights are renamed to experts.switch_glu.
+
+    JANG-quantised Gemma 4 MoE models store expert weights under the key pattern
+      model.language_model.layers.<n>.switch_mlp.{gate,up,down}_proj.{weight,scales,biases}
+    but mlx-vlm's Gemma4TextModel wraps them inside an Experts → SwitchGLU module
+    whose parameter paths look like
+      language_model.model.layers.<n>.experts.switch_glu.{gate,up,down}_proj.{weight,scales,biases}
+
+    The rename must happen *before* model.load_weights() so that strict=False does
+    not silently discard all expert weights (which would leave the model running on
+    random-initialised experts and produce garbled output).
+    """
+
+    # ------------------------------------------------------------------ helpers
+    @staticmethod
+    def _build_shard(layer: int = 0) -> dict:
+        """Return a minimal dict mimicking one JANG shard for a Gemma 4 layer."""
+        prefix = f"model.language_model.layers.{layer}"
+        keys = [
+            f"{prefix}.switch_mlp.gate_proj.weight",
+            f"{prefix}.switch_mlp.gate_proj.scales",
+            f"{prefix}.switch_mlp.gate_proj.biases",
+            f"{prefix}.switch_mlp.up_proj.weight",
+            f"{prefix}.switch_mlp.up_proj.scales",
+            f"{prefix}.switch_mlp.up_proj.biases",
+            f"{prefix}.switch_mlp.down_proj.weight",
+            f"{prefix}.switch_mlp.down_proj.scales",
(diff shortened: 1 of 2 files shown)


jjang-ai:
Thanks for the PR @zheli — you correctly diagnosed the same bug that was reported in #71 by @relus-cy. The fix is shipped in **v1.3.47** (https://github.com/jjang-ai/mlxstudio/releases/tag/v1.3.47).

The shipped fix lives in `vmlx_engine/utils/jang_loader.py` `_load_jang_v2_vlm` and applies the same `.switch_mlp.` → `.experts.switch_glu.` shard rename you wrote, gated on:

```python
_vlm_text_mt = config.get("text_config", {}).get(
    "model_type", config.get("model_type", "")
)
_vlm_needs_gemma4_switch_remap = _vlm_text_mt in ("gemma4", "gemma4_text")
```

(falling back to top-level `model_type` so JANG variants whose `text_config.model_type` is missing still get the rename).

Live verified on `Gemma-4-26B-A4B-it-JANG_4M-CRACK` and `JANG_2L-CRACK` — coherent 5-turn output (Paris / Eiffel / 330m / Gustave Eiffel / recall). Plus 13 other JANG models in the v1.3.47 family regression matrix, all green.

Closing as fixed-via-#71 — your diagnosis was correct, just landed via a parallel PR. Thanks again for the contribution. Recorded as **§8e** in the repo's `NO-REGRESSION-CHECKLIST.md` so this guard can't slip back into the narrow form.

PR #56, opened 2026-04-07
fix: Gemma 4 JANG VLM loading — MoE key mismatch + vision float16 overflow

gucasbrg:
## Summary

Two bugs in `_load_jang_v2_vlm()` that break Gemma 4 JANG models:

### Bug 1: MoE expert weights not loaded → garbled text output

`model.sanitize()` renames `model.language_model.*` → `language_model.model.*` but does **not** rename `switch_mlp.*` → `experts.switch_glu.*`. Since `load_weights(strict=False)` silently skips unmatched keys, all 128×30 MoE expert weights stay random-initialized, producing garbage output.

**Fix:** Add `switch_mlp` → `experts.switch_glu` key rename before `load_weights()`.

### Bug 2: Vision produces all `<pad>` tokens → null content for image inputs

Vision tower weights are stored as float16. During attention computation, scores exceed float16 range (±65504), producing `-inf` values. These propagate through `embed_vision` as `NaN`, causing the model to generate only `<pad>` tokens.

**Fix:** Upcast `vision_tower` and `embed_vision` to bfloat16 (range ±3.4e38) after loading.

## Reproduction

```bash
vmlx serve dealignai/Gemma-4-26B-A4B-JANG_4M-CRACK --port 8080

# Bug 1: garbled text
curl http://localhost:8080/v1/chat/completions \
  -d '{"model":"gemma4","messages":[{"role":"user","content":"What is 2+2?"}],"max_tokens":20}'
# Returns random characters instead of "4"

# Bug 2: vision all-pad (after fixing bug 1)
curl http://localhost:8080/v1/chat/completions \
  -d '{"model":"gemma4","messages":[{"role":"user","content":[
    {"type":"text","text":"Describe this image"},
    {"type":"image_url","image_url":{"url":"https://www.baidu.com/img/flexible/logo/pc/result.png"}}
  ]}],"max_tokens":50}'
# Returns content: null (all tokens are <pad>)
```

## Test plan

- [x] Text generation produces correct output after patch 1
- [x] Vision tower output has no `-inf` or `NaN` after patch 2
- [x] Image description returns meaningful content
- [x] Tested on `dealignai/Gemma-4-26B-A4B-JANG_4M-CRACK` with vMLX 1.3.29 on M4 Max 36GB

### DIFF
--- a/vmlx_engine/utils/jang_loader.py
+++ b/vmlx_engine/utils/jang_loader.py
@@ -816,6 +816,19 @@ def _load_jang_v2_vlm(path: Path, jang_cfg: dict, lazy: bool = False):
                             except Exception:
                                 continue
 
+        # Gemma 4 MoE: rename switch_mlp.* -> experts.switch_glu.* to match model
+        # model.sanitize() renames model.language_model -> language_model.model but
+        # does NOT rename switch_mlp -> experts.switch_glu, causing all MoE expert
+        # weights to silently fail loading (strict=False) and stay random-init.
+        _renamed = {}
+        for _k, _v in shard_weights.items():
+            if ".switch_mlp." in _k:
+                _new_k = _k.replace(".switch_mlp.", ".experts.switch_glu.")
+                _renamed[_new_k] = _v
+            else:
+                _renamed[_k] = _v
+        shard_weights = _renamed
+
         model.load_weights(list(shard_weights.items()), strict=False)
         del shard_weights
         gc.collect()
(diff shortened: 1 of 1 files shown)


jjang-ai:
Thanks @gucasbrg — both parts of your PR are already shipped:

**1) `switch_mlp` → `experts.switch_glu` rename** — landed in **v1.3.47** (#71). Same fix you wrote, gated on:

```python
_vlm_text_mt = config.get("text_config", {}).get(
    "model_type", config.get("model_type", "")
)
_vlm_needs_gemma4_switch_remap = _vlm_text_mt in ("gemma4", "gemma4_text")
```

at `vmlx_engine/utils/jang_loader.py:1107` (with the broadened guard so JANG variants whose `text_config.model_type` is missing still get the rename).

**2) Vision tower float16 overflow → bfloat16 upcast** — landed earlier on **2026-04-09** (was already in v1.3.30+). Lives at `vmlx_engine/utils/jang_loader.py:1412`:

```python
if _text_mt == "gemma4_text" and hasattr(model, "vision_tower"):
    try:
        model.vision_tower.set_dtype(mx.bfloat16)
        if hasattr(model, "embed_vision"):
            model.embed_vision.set_dtype(mx.bfloat16)
        logger.info(
            "  Vision tower upcast to bfloat16 (Gemma 4 VLM — avoids "
            "float16 overflow in SigLIP encoder with low-bit JANG quant)"
        )
```

Same root cause you identified — at the 2L/4M JANG quant levels, SigLIP attention scores can exceed float16 range (±65504), `vision_out` collapses to `-inf`, `embed_vision` becomes NaN, and the language model's first decode step lands on `<pad>`. The shipped fix also covers `embed_vision`, which yours doesn't — both modules need the upcast for the full chain.

Verified live on `Gemma-4-26B-A4B-it-JANG_2L-CRACK` and `JANG_4M-CRACK` in the v1.3.47 regression matrix. Coherent 5-turn output, no `<pad>`.

Closing as already-fixed in tree — thanks for the careful diagnosis and the test repro path (it confirmed the same root cause from a second angle).

PR #77, opened 2026-04-14
Guard default repetition penalty flag for older external engines

### LINKED ISSUE #75: model load error: vmlx-engine: error: unrecognized arguments: --default-repetition-penalty 1.10
加载模型的时候遇到这个错误，所有的选择都是使用默认设置，模型是使用hf的命令行工具下载到本地的，日志如下：
请问如何才能去除这个参数--default-repetition-penalty 1.10
[13:26:36.961] $ /Users/fengjikui/.local/bin/vmlx-engine serve /Users/fengjikui/models/dealignai/Gemma-4-31B-JANG_4M-CRACK --host 127.0.0.1 --port 8000 --timeout 300 --max-num-seqs 5 --prefill-batch-size 512 --prefill-step-size 1024 --completion-batch-size 512 --is-mllm --continuous-batching --tool-call-parser gemma4 --enable-auto-tool-choice --reasoning-parser gemma4 --cache-memory-percent 0.1 --use-paged-cache --paged-cache-block-size 64 --max-cache-blocks 500 --enable-block-disk-cache --block-disk-cache-max-gb 10 --stream-interval 1 --max-tokens 32768 --default-temperature 0.70 --default-top-p 0.95 --default-repetition-penalty 1.10 --default-enable-thinking false
[13:26:37.690] usage: vmlx-engine [-h] {serve,bench,bench-detok,convert,info,list,doctor} ...
[13:26:37.690] vmlx-engine: error: unrecognized arguments: --default-repetition-penalty 1.10
**[13:26:37.735] [ERROR] Process exited with code 2: vmlx-engine: error: unrecognized arguments: --default-repetition-penalty 1.10**

Rishirandhawa:
Fixes #75

## Summary
- detect the resolved local engine version before building session launch args
- omit `--default-repetition-penalty` when the app is launching an older external `vmlx-engine` that predates that CLI flag
- log a clear upgrade hint instead of hard-failing session startup
- keep bundled/project-venv engines on the current behavior
- update the settings-flow test mirror to cover the real default repetition penalty and the legacy-engine compatibility path

## Verification
- `npm test -- settings-flow.test.ts`
- `npm run typecheck` *(currently still fails on the pre-existing renderer typing error at `panel/src/renderer/src/components/sessions/SessionConfigForm.tsx:937`)*

### DIFF
--- a/panel/src/main/engine-manager.ts
+++ b/panel/src/main/engine-manager.ts
@@ -101,7 +101,7 @@ export async function checkEngineInstallation(): Promise<EngineInstallation> {
 /**
  * Get version from vmlx-engine binary
  */
-async function getVersionFromBinary(path: string): Promise<string> {
+export async function getVersionFromBinary(path: string): Promise<string> {
   // Get version via Python package metadata (works with editable installs)
   try {
     const { readFileSync } = await import('fs')
@@ -132,6 +132,29 @@ async function getVersionFromBinary(path: string): Promise<string> {
   } catch (_) { /* not supported */ }
 
   return 'unknown'
+}
+
+export function isVersionAtLeast(version: string | undefined, minimum: string): boolean {
+  if (!version) return false
+
+  const parse = (value: string): number[] | null => {
+    const match = value.match(/^(\d+)\.(\d+)\.(\d+)/)
+    if (!match) return null
+    return match.slice(1).map(Number)
+  }
+
+  const current = parse(version)
+  const target = parse(minimum)
+  if (!current || !target) return false
+
+  for (let i = 0; i < Math.max(current.length, target.length); i++) {
+    const a = current[i] ?? 0
+    const b = target[i] ?? 0
+    if (a > b) return true
+    if (a < b) return false
+  }
+
+  return true
 }
 
 /**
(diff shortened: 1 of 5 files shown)


jjang-ai:
Thanks for the external-engine compatibility work here. I reviewed this in the current release-hardening context.

This is not mergeable as-is now because it conflicts with current panel/session code and the generation-default policy has changed: the panel no longer synthesizes hidden server-level sampling defaults, and repetition penalty should come from model generation config / explicit chat or API override rather than an app-enforced fallback.

If older external-engine support is still needed, the next version should rebase onto current `main` and gate only unsupported CLI flags for external engines without changing model-owned sampling behavior. Please keep the tests focused on the exact unsupported flag/version boundary.

jjang-ai:
Closing as superseded/adapted in current main. Credit to @Rishirandhawa for identifying the older external-engine compatibility failure around --default-repetition-penalty.\n\nCurrent fix direction is deliberately different from this patch: we do not synthesize or force a hidden default repetition penalty. The current app leaves server sampler defaults unset, migrates stale generic defaultRepetitionPenalty=110 session values back to 0, and no longer emits default sampler flags from displayed bundle metadata. That also avoids the older external-engine launch failure class because the flag is not emitted by default.\n\nFresh verification on current source:\n- panel settings-flow focused generation/default tests: 16 passed\n- VLM JIT proxy tests for #154 class: 4 passed\n- release manifest + JANG compat contracts: 21 passed\n\nLeaving the provenance in the PR intake ledger and CHANGELOG credit.

PR #90, opened 2026-04-17
feat: opt-in chunked prefill for hybrid SSM models via VMLX_ALLOW_HYBRID_CHUNKED_PREFILL

### LINKED ISSUE #89: Metal OOM (>72 GB single-buffer) on text-only prompts >~34K tokens with hybrid SSM models (Qwen3.5)
## Environment

- Hardware: Apple M3 Max, 128 GB unified memory
- OS: macOS (Metal single-buffer cap ~72 GiB / 77,309,411,328 bytes)
- vmlx: 1.3.61
- mlx-lm: 0.31.2
- Model: `mlx-community/Huihui-Qwen3.5-27B-Claude-4.6-Opus-abliterated-4bit` (hybrid: GatedDeltaNet + attention, 48 layers)
- Server flags include `--use-paged-cache --continuous-batching --prefill-step-size 512 --prefill-batch-size 1 --kv-cache-quantization q8`

## Bug

Long text-only prompts (>~34K tokens) crash with:

```
libc++abi: terminating due to uncaught exception of type std::runtime_error:
[METAL] Command buffer execution failed: Insufficient Memory
(00000008:kIOGPUCommandBufferCallbackErrorOutOfMemory)
```

A single `attention_scores` allocation of `(1, num_heads, seq_len, seq_len) * 2 bytes` (fp16) at seq_len ≈ 48 000 is ~147 GiB, which exceeds Metal's per-buffer cap of ~72 GiB and OOMs instantly — even though total unified memory is 128 GiB.

## Root cause

`vmlx_engine/mllm_batch_generator.py` unconditionally guards hybrid SSM models against chunked prefill:

- L1428 (fast path): `if not has_images and not self._is_hybrid:` — forces hybrid models through the full model (`self.model(...)`) rather than `language_model(...)` with chunked prefill
- L1444 (chunked path): `if not has_images and seq_len > self.prefill_step_size * 2 and not self._is_hybrid:` — same guard blocks chunked prefill for long prompts

Both paths end up computing full-sequence attention scores for hybrid models, which is what triggers the single-buffer OOM.

## Safety analysis (Qwen3.5)

Chunked prefill is safe for Qwen3.5 hybrid (GatedDeltaNet + attention):

- `mlx_lm/models/qwen3_5.py` uses `create_attention_mask(h, cache[fa_idx])` and `create_ssm_mask(h, cache[ssm_idx])` — both delegate to `cache.make_mask(N)`, which is cache-aware and produces correct masks for any chunk size.
- Both `KVCache` and `ArraysCache` (SSM recurrent state) expose proper `state` properties that carry across chunks.
- vmlx already calls `mx.eval(cache states)` between chunks, materializing KV pairs and SSM recurrent state before the next chunk runs.

Other hybrid architectures have not been verified in mlx_lm's mask creation code, so removing the guards unconditionally is too aggressive.

## Repro

1. Serve Qwen3.5 27B via `vmlx serve ... --prefill-step-size 512 --use-paged-cache --continuous-batching` on M3 Max 128 GB.
2. Send a single request with a ~48K-token text-only prompt.
3. Server crashes with the Metal OOM above, before generating any tokens.

## Proposed fix

Opt-in env var `VMLX_ALLOW_HYBRID_CHUNKED_PREFILL=1` that lets hybrid models take the chunked prefill path on text-only requests. Default remains OFF — no behavior change for existing users.

PR: #<fill-in-after>

Happy to expand verification or adjust the gating strategy (e.g. model-name allowlist, architecture introspection) if preferred.

st-adam:
Fixes #89.

## Summary

Adds an opt-in env var `VMLX_ALLOW_HYBRID_CHUNKED_PREFILL=1` that allows hybrid SSM models (GatedDeltaNet/Mamba + attention) to use chunked prefill on text-only requests. Default remains OFF — no behavior change for existing users.

## Why

Without chunked prefill, hybrid models run full-sequence attention, which allocates `(1, num_heads, seq_len, seq_len) * 2 bytes` of attention scores. On M3 Max 128 GB this exceeds Metal's ~72 GiB per-buffer cap at seq_len ≈ 34K (Qwen3.5 27B hits `kIOGPUCommandBufferCallbackErrorOutOfMemory` around seq_len ≈ 48K).

See #89 for full repro + memory math.

## Changes

One file, `vmlx_engine/mllm_batch_generator.py`:

1. `import os`.
2. In `MLLMBatchGenerator.__init__`, read `VMLX_ALLOW_HYBRID_CHUNKED_PREFILL` into `self._allow_hybrid_chunked_prefill` and log an INFO line when enabled on a hybrid model.
3. Both prefill guards (L1428 fast path, L1444 chunked path) now skip hybrid only when the flag is NOT set. Diff stat: `1 file changed, 35 insertions(+), 5 deletions(-)`.

## Safety

Chunked prefill is safe for Qwen3.5 hybrid:

- `mlx_lm/models/qwen3_5.py` uses cache-aware `create_attention_mask` / `create_ssm_mask` that delegate to `cache.make_mask(N)` — correct for any chunk size.
- Both `KVCache` and `ArraysCache` expose `state` properties that carry across chunks.
- vmlx already `mx.eval`s cache state between chunks.

Other hybrid architectures are unverified — that's why this is opt-in rather than an unconditional guard removal.

## Test plan

- [x] Syntax: `python3 -m py_compile vmlx_engine/mllm_batch_generator.py`
- [x] Default (flag unset): hybrid models go through single-shot path unchanged
- [x] Flag set: Qwen3.5 27B with a 48K-token text-only prompt completes without Metal OOM; prefill chunks at `prefill_step_size` (512)
- [x] Flag set: VLM requests with images still run in one shot (vision encoder needs full sequence)
- [x] Non-hybrid models unaffected (`self._is_hybrid=False` short-circuits regardless of flag)

Happy to adjust the flag name, default, or gating strategy (e.g. allowlist by model architecture) based on review.

### DIFF
--- a/vmlx_engine/mllm_batch_generator.py
+++ b/vmlx_engine/mllm_batch_generator.py
@@ -98,6 +98,7 @@ HELPER FUNCTIONS
 """
 
 import logging
+import os
 import time
 from collections import OrderedDict
 from dataclasses import dataclass, field
@@ -1040,6 +1041,28 @@ class MLLMBatchGenerator:
             and len(self._hybrid_kv_positions) < self._hybrid_num_layers
         )
 
+        # Opt-in: enable chunked prefill for hybrid SSM models on text-only
+        # requests. Default OFF preserves the conservative single-shot path.
+        # When enabled, long text-only prompts (>~34K tokens) avoid the Metal
+        # single-buffer OOM from full-sequence attention scores:
+        #   (1, num_heads, seq_len, seq_len) * 2 bytes exceeds 72 GB at
+        #   seq_len >~ 34K on M3 Max 128 GB (72 GB = Metal's per-buffer cap).
+        # Verified safe for Qwen3.5 (GatedDeltaNet + attention): its
+        # create_attention_mask / create_ssm_mask in mlx_lm are cache-aware
+        # and produce correct masks for any chunk size, and both KVCache and
+        # ArraysCache carry state across chunks via mx.eval() at each step.
+        # Other hybrid architectures are not verified — hence opt-in.
+        self._allow_hybrid_chunked_prefill: bool = os.environ.get(
+            "VMLX_ALLOW_HYBRID_CHUNKED_PREFILL", "0"
+        ).lower() in ("1", "true", "yes", "on")
+        if self._is_hybrid and self._allow_hybrid_chunked_prefill:
+            logger.info(
+                "MLLMBatchGenerator: VMLX_ALLOW_HYBRID_CHUNKED_PREFILL=1 — "
+                "chunked prefill enabled for this hybrid SSM model on "
+                "text-only requests. Verified safe for Qwen3.5; other hybrid "
+                "architectures are unverified."
+            )
+
         # Vision embedding cache for repeated images
         self.vision_cache = VisionEmbeddingCache(
             max_pixel_entries=vision_cache_size,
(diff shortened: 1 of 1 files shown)


st-adam:
Thanks @jjang-ai — I see v1.3.62 (2026-04-18 CHANGELOG) shipped the `VMLX_ALLOW_HYBRID_CHUNKED_PREFILL=1` opt-in with the same gating logic I proposed here (`self._is_hybrid and not _allow_hybrid_chunked` around both the fast-path and chunked-prefill branches). Closing this PR as superseded by that release. Happy to open targeted follow-ups for the other hybrid-model issues I've been working on (SSM companion persistence #91, and two new scheduler bugs I've since isolated).

st-adam:
Closing: superseded by v1.3.62 (same env-var opt-in, same gating logic).

jjang-ai:
Hey Adam, I'd love any further reports - I'd go out of my way to fix it myself but I have so many projects. Any kind of further issue reports / specifics are always appreciated!

st-adam:
Many thanks Jinho, happy to help!

PR #93, opened 2026-04-19
fix(scheduler): guard PLD speculative decode on non-MLLM BatchGenerator (#92)

### LINKED ISSUE #92: PLD speculative decode crashes on non-MLLM BatchGenerator, corrupts batch state and wipes paged cache
## Summary

`Scheduler._try_speculative_decode()` assumes the BatchGenerator exposes `MLLMBatchGenerator`'s API (`.active_batch` for forward-logprobs peek; `remove(..., return_prompt_caches=True)` returning cache objects that expose `.is_trimmable()`). On text-only / non-MLLM paths the generator is mlx-lm's plain `BatchGenerator`, which has neither — PLD-spec crashes, the `finally`-block's emergency re-insert pushes the malformed cache back into the batch, and the next `step()` raises `<class 'list'> does not yet support batching with history`. The scheduler then `clear()`s the entire paged cache and forces a full prefill on the recovery retry.

## Environment

- vmlx `1.3.62` (and earlier — same structure back to when PLD landed in #26)
- macOS 14+, Apple Silicon
- Any non-MLLM model (e.g. text-only Qwen3 invoked without the VLM wrapper)
- `--enable-pld` on the server, `temperature <= self._pld_spec_max_temp` on the request

## Repro sketch

1. Start server with `--enable-pld` on a pure text model (no `vision_config`).
2. Send any request with repetitive-enough output that PLD finds drafts.
3. First `_try_speculative_decode` call hits `self.batch_generator.active_batch` — `AttributeError` — and falls into the try/except → finally path.
4. Next batch `step()` then raises `<class 'list'> does not yet support batching with history` from the cache-merge path.
5. Recovery path clears the paged cache (and prefix cache) and re-prefills from scratch.

## Why it matters

- Even on pure-attention text models `_try_speculative_decode` runs for every request once `--enable-pld` is set, so any single PLD-enabled server hitting a non-MLLM code path corrupts itself in the first few tokens.
- The recovery clears every cached prefix for every in-flight and queued request, so the cost is not localized to the crashing request — next turn of a 20K-token conversation pays full re-prefill.

## Proposed fix

Short-circuit BEFORE `remove()` with a single capability check:

```python
if not hasattr(self.batch_generator, "active_batch"):
    return []
```

Place it right after the existing `temp > self._pld_spec_max_temp` gate so no batch state is touched on non-MLLM paths. The retrospective prompt-lookup analyzer in `prompt_lookup.py` still runs inline on every decode step, so PLD hit-rate telemetry / `theoretical_speedup` stats stay accurate — only the batched verify-and-reinsert cycle is gated off.

Happy to open a PR with the guard + a unit test that proves `_try_speculative_decode` returns `[]` without touching the generator when `.active_batch` is absent.

st-adam:
## Summary

Fixes #92.

`_try_speculative_decode` assumes `MLLMBatchGenerator`'s API surface — specifically `.active_batch` for the forward_logprobs peek and `remove(..., return_prompt_caches=True)` returning cache objects with `.is_trimmable()`. mlx-lm's text-only `BatchGenerator` has neither, so a non-MLLM deployment hits:

1. `AttributeError` at `self.batch_generator.active_batch`
2. the finally-block's emergency re-insert corrupts batch state
3. the *next* `step()` dies with `<class 'list'> does not yet support batching with history` inside `_patched_merge_caches`, which triggers a full paged-cache wipe on recovery

This PR adds a single-point `hasattr(self.batch_generator, \"active_batch\")` gate right after the temperature check and **before any batch-state mutation**, so non-MLLM deployments can't corrupt themselves.

Retrospective prompt-lookup telemetry in `prompt_lookup.py` continues to run on every decode step, so PLD hit-rate reporting is unaffected — only the batched verify-and-reinsert cycle is gated off on non-MLLM generators.

## Why a capability check instead of an isinstance check

`hasattr(..., \"active_batch\")` is stable under subclassing and doesn't couple scheduler code to import order of the MLLM module. The two things PLD-spec actually *needs* are `active_batch` + `is_trimmable()` prompt caches; the `active_batch` attribute is a reliable proxy because both are specific to `MLLMBatchGenerator`.

## Test plan

New: `tests/test_pld_non_mllm_guard.py` — 3 cases
- [x] `test_returns_empty_on_non_mllm_generator` — non-MLLM `BatchGenerator` (no `.active_batch`) returns `[]`, `remove()` / `insert()` never called
- [x] `test_does_not_count_as_spec_attempt` — guard does not increment `_pld_spec_attempts`, so [PLD:*] summaries don't report phantom attempts
- [x] `test_mllm_generator_not_short_circuited_by_guard` — sanity check: `hasattr == True` path reaches the downstream uid lookup (guard doesn't over-fire)

Regressions:
- [x] All 51 existing tests in `tests/test_speculative.py` still pass
- [x] Tested locally on Qwen3.6-35B-A3B-Abliterated-Heretic on M3 Max 128GB; PLD hit-rate telemetry unchanged, no more `<class 'list'> does not yet support batching` crashes

## Risk / scope

- Single 3-line edit in `_try_speculative_decode` + explanatory comment
- No changes to behavior on the MLLM path (the guard is a no-op there)
- No new dependencies

### DIFF
--- a/tests/test_pld_non_mllm_guard.py
+++ b/tests/test_pld_non_mllm_guard.py
@@ -1,0 +1,108 @@
+# SPDX-License-Identifier: Apache-2.0
+"""
+Regression tests for vmlx#92 — PLD speculative decode must short-circuit
+when the BatchGenerator isn't MLLMBatchGenerator.
+
+Before the guard, `_try_speculative_decode` reached `self.batch_generator.
+active_batch`, raised AttributeError, and the finally-block's emergency
+re-insert then corrupted the batch so the next `step()` died with
+`<class 'list'> does not yet support batching with history` — triggering
+a full paged-cache wipe on recovery. Guarding BEFORE `remove()` keeps the
+batch state clean and lets the retrospective prompt-lookup analyzer
+continue to run normally.
+"""
+
+from types import SimpleNamespace
+from unittest.mock import MagicMock
+
+import pytest
+
+
+class TestPLDNonMLLMGuard:
+    """vmlx#92 — short-circuit PLD-spec when active_batch is absent."""
+
+    def _make_scheduler_shim(self, batch_generator):
+        """Build the minimum attribute surface `_try_speculative_decode`
+        touches before the guard fires: temperature gate, n-gram cache,
+        uid map, spec-attempt counter, batch_generator."""
+        from vmlx_engine.scheduler import Scheduler
+
+        sched = Scheduler.__new__(Scheduler)
+        sched._pld_spec_max_temp = 0.8
+        sched._pld_ngram_indices = {}
+        sched._pld_spec_attempts = 0
+        sched.request_id_to_uid = {"req-1": 42}
+        sched.batch_generator = batch_generator
+        return sched
+
+    def _make_request(self):
+        """Request stub carrying the minimum fields the early gates read."""
+        req = SimpleNamespace()
+        req.sampling_params = SimpleNamespace(
+            temperature=0.0,
+            max_tokens=128,
+        )
+        req.prompt_token_ids = [10, 11, 12, 13, 14, 15, 16, 17, 18, 19]
+        # Force the n-gram lookup to find drafts: repeat a 3-gram so the
+        # n-gram index surfaces the continuation, and ensure we're past the
+        # `remaining <= 1` gate.
+        req.output_token_ids = [20, 21, 22, 23, 20, 21, 22]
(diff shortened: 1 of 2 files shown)


st-adam:
Closing — the fix landed directly on main as commit \`8fcada93\` in v1.3.63+, with the exact guard I proposed. Three regression tests ship in \`tests/test_vl_video_regression.py::TestVmlx92PldGuardOnNonMllm\` pinning the contract. Thanks!

PR #104, opened 2026-04-24
fix(scheduler): gate deferred SSM re-derive on empty waiting queue (vmlx#103)

### LINKED ISSUE #103: Scheduler: deferred SSM re-derive can starve queued requests (missing `not self.waiting` guard)
## Summary

The deferred-SSM-re-derive loop at `vmlx_engine/scheduler.py:4605-4613` runs a re-derive prefill pass whenever `self.running` is empty, without also checking `self.waiting`. A request that is sitting in the admission queue (submitted but not yet promoted to `running`) will therefore eat a full re-derive step's worth of TTFT latency for every pending item in `_ssm_rederive_queue`.

## Current code

```python
if (
    self._is_hybrid
    and not self.running
    and hasattr(self, "_ssm_rederive_queue")
    and self._ssm_rederive_queue
    and self._ssm_state_cache is not None
):
    # Process ONE task per step to avoid long GPU stalls
    tokens, prompt_len, orig_request_id = self._ssm_rederive_queue.pop(0)
    ...
```

## Problem

The docstring says "ONLY when the scheduler is idle (no running requests)" — but `running` is just the currently-generating set. `waiting` holds FCFS-queued requests that haven't been admitted yet (see `self.waiting: deque[Request]` at L268, admission at L2768). When a new request arrives on an otherwise-quiet hybrid engine, step order is:

1. User request lands → enqueued in `self.waiting`
2. Scheduler step begins, `self.running` is empty
3. Gate passes → re-derive runs a full prefill pass (potentially hundreds of ms on a 12k-token prompt)
4. New request finally admitted on the *next* step

With `_ssm_rederive_queue_max = 20` (`mllm_batch_generator.py:1005-1006`), a freshly-woken engine with a full queue can silently burn 20 re-derive passes before servicing the first interactive request.

## Reproduction

On a hybrid model (e.g. Qwen3.6 A3B, Qwen3.5 GatedDeltaNet variants):

1. Fill `_ssm_rederive_queue` with several items (close N prior hybrid turns with `VMLX_ENABLE_SSM_PREFIX_RESUME=1` and `gen_prompt_len > 0`)
2. Let `running` drain to empty
3. Submit a new request
4. Observe TTFT spike — measurable even with one queued re-derive on a 8k-token prompt

## Proposed fix

One-line gate addition:

```python
if (
    self._is_hybrid
    and not self.running
    and not self.waiting          # ← add this
    and hasattr(self, "_ssm_rederive_queue")
    and self._ssm_rederive_queue
    and self._ssm_state_cache is not None
):
    ...
```

Re-derive is already opportunistic ("pays a cache miss now to save time on the next conversation turn") — deferring it by one step when there is a waiting user is strictly better for interactive UX. Throughput is unaffected: the queue still drains whenever the engine goes fully idle, which is exactly what the deferred path was designed for.

## Impact

- Latency win: eliminates N × re-derive TTFT penalty when queue is full and a new request arrives on an idle engine
- No functional change: re-derive still runs, just never ahead of a queued user request
- Zero risk: the change only narrows the condition under which an optional background task runs

## Version info

- vmlx 1.3.86 (upstream/main tip `fa1fdb8`)
- Affected models: all hybrid SSM families (qwen3_next, qwen3_5, qwen3_6 A3B variants)

st-adam:
## Summary

Closes #103. The deferred SSM re-derive loop at `vmlx_engine/scheduler.py:4605-4611` only guarded on `not self.running`, so any request sitting in `self.waiting` (FCFS admission queue) paid a full re-derive prefill's worth of TTFT for every queued item in `_ssm_rederive_queue` before being admitted to generation.

This one-line fix adds `and not self.waiting` to the gate so re-derive runs only when the engine is *fully* idle — no generating requests **and** no queued admissions.

## Why this is safe

Re-derive is opportunistic by design — it trades current-turn latency for next-turn cache warmth on hybrid SSM models. The change only **narrows** the condition under which an optional background task runs; it never blocks a user-visible operation and never changes generation output.

- Throughput on a busy engine: unchanged (queue never met the gate)
- Throughput on a drained engine: unchanged (queue drains as before, just one step later when a new request happens to arrive)
- Latency for a request arriving on an idle engine with queued re-derives: eliminates N × re-derive TTFT penalty

## Test plan

- [x] Read verified against `upstream/main` tip `fa1fdb8` — `self.waiting` confirmed as `deque[Request]` at `scheduler.py:268` with admission at `scheduler.py:2768`
- [ ] Reproduction on hybrid model (Qwen3.6 A3B or Qwen3.5 GatedDeltaNet): warm `_ssm_rederive_queue` with ≥5 items, drain `running`, submit new request, measure TTFT before/after patch
- [ ] Existing hybrid SSM regression tests (`tests/test_vl_video_regression.py`, `tests/test_jit_toggle.py`) pass unchanged

## Notes for review

- Follow-up candidate (not in this PR): reconsider `_ssm_rederive_queue_max = 20` — with this gate the queue still drains, but an unbounded queue can hold memory proportional to prompt length × queue size. Current cap looks fine.
- Related recent work: v1.3.84 fixed a different re-derive bug (broadcast-shape mismatch on auto-chunked prefill). This PR is independent and composes cleanly.

### DIFF
--- a/vmlx_engine/scheduler.py
+++ b/vmlx_engine/scheduler.py
@@ -4594,17 +4594,20 @@ class Scheduler:
         # ── Deferred SSM re-derive (idle-time processing) ──
         # For thinking models (gen_prompt_len > 0), the SSM companion store
         # queues a re-derive task instead of skipping entirely. We run the
-        # re-derive here ONLY when the scheduler is idle (no running
-        # requests) — the forward pass uses the Metal GPU so it can't
-        # overlap with active generation. The re-derive runs a separate
-        # prefill pass on just the prompt tokens (no thinking/output
-        # contamination) and stores the clean SSM state for future prefix
-        # cache hits. This means the CURRENT conversation pays full
-        # re-prefill on each turn, but the NEXT conversation with the
+        # re-derive here ONLY when the scheduler is fully idle (no running
+        # AND no queued-waiting requests) — the forward pass uses the Metal
+        # GPU so it can't overlap with active generation, and `self.waiting`
+        # must also be empty so we don't steal TTFT from interactive requests
+        # that just landed on the admission queue (vmlx#103). The re-derive
+        # runs a separate prefill pass on just the prompt tokens (no
+        # thinking/output contamination) and stores the clean SSM state for
+        # future prefix cache hits. This means the CURRENT conversation pays
+        # full re-prefill on each turn, but the NEXT conversation with the
         # same prompt prefix gets an instant KV+SSM cache hit.
         if (
             self._is_hybrid
             and not self.running
+            and not self.waiting
             and hasattr(self, "_ssm_rederive_queue")
             and self._ssm_rederive_queue
             and self._ssm_state_cache is not None


st-adam:
Closing — superseded by a stricter upstream gate that landed in v1.3.93.

The deferred SSM re-derive cleanup path (around `scheduler.py:4616`) now gates on three conditions instead of the two in this PR:

```python
_has_unprocessed = bool(getattr(self, "unprocessed_requests", []))
_has_waiting = bool(getattr(self, "waiting", []))
if (
    self._is_hybrid
    and not self.running
    and not _has_waiting
    and not _has_unprocessed
    ...
):
```

The third condition (`unprocessed_requests`) is the right addition — it covers tasks that have been admitted but haven't ticked yet, which can starve identically to `waiting`-queue items. My PR missed that.

Closing this; the upstream shape covers the case better.

PR #106, opened 2026-04-24
chore(ssm): drop redundant `mx.array()` wrap before `mx.contiguous` (vmlx#105)

### LINKED ISSUE #105: SSM capture: `mx.contiguous(mx.array(a))` wraps an existing mx.array redundantly
## Summary

Three sites in the SSM-companion cache-clone path do `mx.contiguous(mx.array(a))` where `a` is already an `mx.array` (it comes straight out of `c.cache: List[mx.array]`). `mx.array(x)` is a no-op when `x` is already an mx.array, so the wrap adds no semantics — `mx.contiguous(a)` is equivalent and clearer.

## Call sites (`upstream/main` @ `fa1fdb8`)

- `vmlx_engine/scheduler.py:4633` — deferred SSM re-derive, clean-cache SSM layer extraction
- `vmlx_engine/mllm_batch_generator.py:2311` — async MLLM re-derive watcher
- `vmlx_engine/mllm_batch_generator.py:2938` — MLLM immediate one-shot re-derive

All three are the same pattern, with the same inline comment intent ("Ensure MLX arrays are fully materialized copies") — the materialization comes from `mx.contiguous`, not from the outer `mx.array` wrap.

## Proposed change

```python
# before
cloned.cache = [
    mx.contiguous(mx.array(a)) if a is not None else None
    for a in c.cache
]
# after
cloned.cache = [
    mx.contiguous(a) if a is not None else None
    for a in c.cache
]
```

## Correctness argument

- `a` comes from `c.cache` which is typed `List[mx.array]` — confirmed by the `isinstance(c.cache, list)` guard and the companion cache API
- `mx.array(existing_mx_array)` in mlx-core returns the same array object (the Python binding treats an existing `mx.array` input as identity — see mlx `mlx/core/array.cpp` constructor dispatch)
- Behavior is therefore bit-identical: `mx.contiguous(mx.array(a))` → `mx.contiguous(a)`
- The `deepcopy(c)` above already handles the cache-wrapper clone; the list comprehension overwrites `cloned.cache` regardless, so the outer `mx.array` was never defensive against aliasing anyway

## Impact

Pure cleanup — no functional change. One redundant function-call site removed per iteration × (SSM layer count × re-derive frequency). Not a perf win worth measuring; just hygiene and readability.

## Version info

- vmlx 1.3.86 (`fa1fdb8`)
- mlx 0.31.2

## Related work

Companion PR will follow immediately after this issue.

st-adam:
## Summary

Closes #105. Three SSM-companion cache-clone sites wrap `a in c.cache` with `mx.array()` before passing it to `mx.contiguous()`. Since `c.cache` is typed `List[mx.array]` and `mx.array(existing_mx_array)` is a no-op, the outer wrap is pure noise. This PR simplifies all three sites to `mx.contiguous(a) if a is not None else None`.

## Changed sites

| File | Line | Context |
|---|---|---|
| `vmlx_engine/scheduler.py` | 4633 | Deferred SSM re-derive (LLM scheduler) |
| `vmlx_engine/mllm_batch_generator.py` | 2311 | Async MLLM re-derive watcher |
| `vmlx_engine/mllm_batch_generator.py` | 2938 | MLLM immediate one-shot re-derive |

## Correctness

Zero behavior change. `mx.array(x)` returns `x` when `x` is already an `mx.array` (identity dispatch in mlx-core's Python binding constructor). Materialization comes from `mx.contiguous()` itself — the outer wrap was never doing anything. The `isinstance(c.cache, list)` guard above each call site confirms the type assumption.

## Test plan

- [x] Verified against `upstream/main` tip `fa1fdb8`
- [ ] Existing hybrid SSM regression tests pass unchanged (`tests/test_vl_video_regression.py`, `tests/test_jit_toggle.py`)
- [ ] Smoke test with Qwen3.6 A3B / Qwen3.5 GatedDeltaNet — clean-cache clone path still produces usable companion caches

## Notes for review

Small hygiene patch intended to pair with the neighboring #104 starvation-gate fix — both touch the same re-derive code paths. Happy to squash or merge in either order.

### DIFF
--- a/vmlx_engine/mllm_batch_generator.py
+++ b/vmlx_engine/mllm_batch_generator.py
@@ -2308,7 +2308,7 @@ class MLLMBatchGenerator:
                                     cloned = deepcopy(c)
                                     # Ensure MLX arrays are fully materialized copies
                                     cloned.cache = [
-                                        mx.contiguous(mx.array(a)) if a is not None else None
+                                        mx.contiguous(a) if a is not None else None
                                         for a in c.cache
                                     ]
                                     ssm_layers.append(cloned)
@@ -2935,7 +2935,7 @@ class MLLMBatchGenerator:
                     from copy import deepcopy
                     cloned = deepcopy(c)
                     cloned.cache = [
-                        mx.contiguous(mx.array(a)) if a is not None else None
+                        mx.contiguous(a) if a is not None else None
                         for a in c.cache
                     ]
                     ssm_layers.append(cloned)
--- a/vmlx_engine/scheduler.py
+++ b/vmlx_engine/scheduler.py
@@ -4630,9 +4630,7 @@ class Scheduler:
 
                                 cloned = deepcopy(c)
                                 cloned.cache = [
-                                    mx.contiguous(mx.array(a))
-                                    if a is not None
-                                    else None
+                                    mx.contiguous(a) if a is not None else None
                                     for a in c.cache
                                 ]
                                 ssm_layers.append(cloned)


st-adam:
Closing — landed upstream verbatim in v1.3.93.

The three SSM cache-clone sites (`scheduler.py`, `mllm_batch_generator.py` × 2) now use `mx.contiguous(a)` directly without the redundant `mx.array()` wrap, matching this PR exactly. Verified against `v1.3.93` source (commit `eec97d3 v1.3.93: canonical engine source — restore from /tmp build dir to git`) — the relevant changes mention `jjang-ai/vmlx#105: removed redundant mx.array() wrap in mamba cache clones`.

Closing.

Issue #110, opened 2026-04-24
SSM companion disk persistence: optional L2 write-through on SSMCompanionCache (follow-up to #91)


st-adam:
## Background

Follow-up to my [#91 comment](https://github.com/jjang-ai/vmlx/issues/91#issuecomment-4276546217) where I flagged this direction and offered to hold until the in-memory `fetch_longest_prefix` path matured. That path has now shipped (v1.3.66), been hardened (v1.3.76 `is_complete` gating, v1.3.78 slice fix, v1.3.84 broadcast fix), and proven stable across v1.3.66 → v1.3.86. Filing the formal proposal so this lives somewhere findable rather than in a closed-issue comment.

## Gap

The in-memory `SSMCompanionCache` (`vmlx_engine/utils/ssm_companion_cache.py`) is lost on every engine restart. For workloads with stable system prompts (my case: Home Assistant voice + general chat against a fixed 5k-token system prompt, shared across sessions), every cold start pays full prefill on that prompt — even though the engine computed the same clean SSM state dozens of times on prior runs.

Paged KV blocks *do* have a disk layer (`block_disk_store.py`, landed upstream), but it's orthogonal — SSM companion is a separate store with its own key/value shape.

## Proposal

**Strictly additive, gated off by default**, mirroring the rollout pattern of `VMLX_ENABLE_SSM_PREFIX_RESUME`:

1. **`utils/ssm_companion_disk_store.py`** — reference backend
   - Storage: safetensors for the SSM layer `mx.array` tensors + JSON sidecar for metadata (`{model_key, num_tokens, prefix_hash, is_complete, layer_schema}`)
   - Keyed on SHA-256 of tokens (same hash the in-memory store already computes)
   - Atomic write via rename; corruption-on-read skipped with a WARN and treated as a miss

2. **`SSMCompanionCache(disk_store=...)`** — optional kwarg
   - `None` (default) → bit-identical behavior to today
   - Non-None → write-through from `store()`, read-through fallback from `fetch()` and `fetch_longest_prefix()`
   - In-memory LRU stays the hot path; disk only touched on in-memory miss

3. **New env var**: `VMLX_SSM_COMPANION_DISK_CACHE_DIR=<path>` wired at scheduler / mllm_batch_generator construction. Unset → disabled.

4. **Eviction / size cap**: `VMLX_SSM_COMPANION_DISK_MAX_GB` with LRU eviction driven by sidecar mtime. Default 10 GB as a rough starting point.

[...]

PR #111, opened 2026-04-24
feat(ssm): optional L2 disk persistence for SSM companion cache (vmlx#110)

st-adam:
## Summary

Addresses #110 with a concrete proposal. **Still welcoming opinions on the 4 open questions below before this is merge-ready** — treating this PR as a sketch + running reference implementation to anchor the conversation rather than a "please merge" request. Happy to rework any of it.

Adds an optional L2 disk store so the SSM companion cache survives process restarts. Paged KV blocks already persist via `block_disk_store.py`, so a cold-start workload with a stable system prompt currently pays full prefill on the SSM side even when the KV side hits from disk. This PR closes that gap while leaving the pure-memory path bit-identical when the env var is unset.

### New: `vmlx_engine/utils/ssm_companion_disk_store.py` (361 lines)

- `SSMCompanionDiskStore(cache_dir, max_size_gb=10.0)` — safetensors + JSON sidecar layout, atomic rename on write, LRU eviction by mtime
- Writes batched to a single background `ThreadPoolExecutor` so the scheduler hot path never blocks on fsync
- Reads inline (`mx.load` is mmap-backed and cheap); corrupt sidecar / missing binary treated as a miss
- Orphaned tmp files cleaned up on init
- `build_from_env()` module-level helper returning `None` unless `VMLX_SSM_COMPANION_DISK_DIR` is set — keeps the construction sites terse and the surface area auditable

### Modified: `vmlx_engine/utils/ssm_companion_cache.py` (+49 / -3)

- `__init__` gains optional `disk_store: Optional[Any] = None`; `None` keeps legacy behaviour unchanged
- `store()` adds L2 write-through after L1 eviction
- `fetch()` adds L2 read-through on L1 miss, promoting the disk hit back into the L1 LRU **and** registering it in `_length_index` so the same query's subsequent `fetch_longest_prefix` calls stay hot
- `SSMCompanionDiskStore` re-exported through the cache module so the stable import path exposes both classes

### Modified: `vmlx_engine/scheduler.py` + `vmlx_engine/mllm_batch_generator.py` (+8 / -1 each)

Both `HybridSSMStateCache` construction sites now call `build_from_env()` for the `disk_store` kwarg.

## Env vars (mirror the `VMLX_ENABLE_SSM_PREFIX_RESUME` rollout pattern)

- `VMLX_SSM_COMPANION_DISK_DIR` — root path. Unset ⇒ disabled.
- `VMLX_SSM_COMPANION_DISK_MAX_GB` — byte cap (default 10.0).

## Compatibility / risk

- Strictly additive. `disk_store=None` is bit-identical to pre-PR behaviour — zero overhead on the hot path when the env var is unset
- `store()` / `fetch()` / `fetch_longest_prefix()` return shapes unchanged
- `is_complete` flag survives the round trip via the sidecar JSON, so v1.3.76's gpl-contamination gate keeps working across restart
- All disk I/O exceptions are caught; failures log at DEBUG or WARNING and the L1 path is never impacted

## Open design questions (from #110)

Still actively seeking your take on these before we land:

1. **Config surface** — env var only (as implemented), or also a CLI flag `--ssm-companion-disk-cache-dir`? I went env-var-only to match the `VMLX_ENABLE_SSM_PREFIX_RESUME` precedent but happy to add the CLI side.
2. **Serialisation format** — safetensors + JSON sidecar (this PR), or align with whatever `block_disk_store.py` uses for consistency? I haven't audited the block store format; open to matching it if you'd prefer a single on-disk convention.
3. **Eviction policy** — LRU on sidecar mtime (this PR), or something coupled to the L1 LRU? Coupling would require callbacks from `SSMCompanionCache.fetch/store` into the disk store; I stayed decoupled for simplicity.
4. **Split the PR?** — I can carve this into (a) disk-store module, (b) cache-layer integration, (c) scheduler + mllm_batch_generator wiring if you'd rather review in pieces.

### Completed
- [x] Python syntax check on all four touched files
- [x] `SSMCompanionCache(disk_store=None)` smoke test — legacy path unchanged, `fetch()` on empty → None
- [x] `build_from_env()` returns None when env var unset, builds correctly when set with custom `VMLX_SSM_COMPANION_DISK_MAX_GB`
- [x] Re-export identity: `SSMCompanionDiskStore is ssm_companion_cache.SSMCompanionDiskStore`

### Needs owner environment
- [ ] End-to-end on Qwen3.6 35B A3B / Qwen3.5 GatedDeltaNet: warm SSM companion → restart engine → confirm next request hits L2 (should observe the "cached=P / remaining=0" log from v1.3.66 for the exact same prompt that previously forced full prefill)
- [ ] Eviction under load: exceed `VMLX_SSM_COMPANION_DISK_MAX_GB`, observe LRU-by-mtime removal of oldest sidecars
- [ ] Crash resilience: kill the engine mid-write, restart, confirm orphan `*.tmp.*` files cleaned up on init and no corrupt reads on fetch
- [ ] Regression pass for existing SSM tests (the ones that pinned v1.3.78 slice fix, v1.3.84 broadcast fix): expectation is zero behavioural change with disk store off, and identical semantics (plus disk persistence) with it on

## Form-4 provenance

I've been running the underlying implementation locally on Qwen3.6 35B A3B (M3 Max 128GB) for ~2 weeks via a standalone patch script — this PR is a clean port against `upstream/main` tip `fa1fdb8` (v1.3.86). Observed cold-start TTFT on a 5k-token system prompt drop from ~3.2s → ~280ms with the disk store on; disk footprint was ~90 MB per Qwen3.6 A3B checkpoint (30 GatedDeltaNet layers × ~3 MB each). Happy to re-run those numbers against current upstream once this is reviewed.

## Version info

- Base: `upstream/main` tip `fa1fdb8` (v1.3.86)
- mlx 0.31.2, mlx-lm 0.31.3

## Related

Orthogonal to my other in-flight PRs (#104, #106, #108) — this one touches different code paths (utils layer + construction, not the scheduler step loop). Any merge order works.

### DIFF
--- a/vmlx_engine/mllm_batch_generator.py
+++ b/vmlx_engine/mllm_batch_generator.py
@@ -994,7 +994,15 @@ class MLLMBatchGenerator:
         # Companion SSM state cache for hybrid models (MambaCache + KVCache).
         # Stores SSM layer states at prompt boundary so hybrid cache HITs can
         # skip the full prefix instead of wasting the KV cache hit.
-        self._ssm_state_cache = HybridSSMStateCache(max_entries=ssm_state_cache_size)
+        # vmlx#110: optional L2 disk store so the companion survives process
+        # restart. Off unless VMLX_SSM_COMPANION_DISK_DIR is set; safe no-op.
+        from .utils.ssm_companion_disk_store import (
+            build_from_env as _build_ssm_disk_store,
+        )
+        self._ssm_state_cache = HybridSSMStateCache(
+            max_entries=ssm_state_cache_size,
+            disk_store=_build_ssm_disk_store(),
+        )
 
         # Async rederive queue for MLLM thinking models. When we capture SSM
         # state post-full-prefill (is_complete=False, gpl-contaminated), we
--- a/vmlx_engine/scheduler.py
+++ b/vmlx_engine/scheduler.py
@@ -356,7 +356,16 @@ class Scheduler:
                 # MLLM scheduler honoured `MLLMSchedulerConfig.ssm_state_cache_size`
                 # but the LLM scheduler hardcoded 50, ignoring user config.
                 _ssm_cache_size = getattr(self.config, "ssm_state_cache_size", 50) or 50
-                self._ssm_state_cache = HybridSSMStateCache(max_entries=_ssm_cache_size)
+                # vmlx#110: optional L2 disk store so the SSM companion
+                # survives process restart. Off unless
+                # VMLX_SSM_COMPANION_DISK_DIR is set; safe no-op otherwise.
+                from .utils.ssm_companion_disk_store import (
+                    build_from_env as _build_ssm_disk_store,
+                )
+                self._ssm_state_cache = HybridSSMStateCache(
+                    max_entries=_ssm_cache_size,
+                    disk_store=_build_ssm_disk_store(),
+                )
                 logger.info(
                     f"Hybrid SSM cache: {len(self._hybrid_kv_positions)}/"
                     f"{self._hybrid_num_layers} KV layers, "
--- a/vmlx_engine/utils/ssm_companion_cache.py
+++ b/vmlx_engine/utils/ssm_companion_cache.py
@@ -134,7 +134,12 @@ class SSMCompanionCache:
     LRU eviction keeps it bounded.
     """
 
-    def __init__(self, max_entries: int = 20, model_key: str = ""):
+    def __init__(
+        self,
+        max_entries: int = 20,
+        model_key: str = "",
+        disk_store: Optional[Any] = None,
+    ):
         if max_entries < 1:
             raise ValueError("max_entries must be >= 1")
         # Internal storage: key -> (states, is_complete) tuple.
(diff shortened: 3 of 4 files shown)


st-adam:
Closing — superseded by `SSMCompanionDiskStore` shipped in v1.3.93.

Upstream shipped `vmlx_engine/utils/ssm_companion_disk_store.py` (520 lines) with a more polished shape than this PR's sketch:

- Two-level fan-out by hash prefix (avoids single-dir file accumulation under high entry counts).
- Off-by-default behind `VMLX_ENABLE_SSM_DISK_CACHE=1` (this PR was always-on once the dir env var was set — the gate is the right call).
- Atomic write via tmp + `os.replace` keeping the `.safetensors` extension (mlx requires it on read).
- LRU eviction by mtime under `VMLX_SSM_DISK_CACHE_MAX_GB` (default 10 GB).
- Corrupt sidecar / load failure → treated as miss (never returns partial result).
- mkdir / permission failure → store inert, returns `None` from `get_disk_store()`, L1 keeps working.
- Tests at `tests/test_ssm_companion_disk_store.py` (10 unit) + `tests/test_ssm_production_matrix.py::T9..T11b` (cross-process, isolation under collision, permission-denied graceful degrade, disk-full simulation).

The upstream env var names differ from this PR's `VMLX_SSM_COMPANION_DISK_*`:
- `VMLX_ENABLE_SSM_DISK_CACHE` (gate, off by default)
- `VMLX_SSM_DISK_CACHE_DIR`
- `VMLX_SSM_DISK_CACHE_MAX_GB`
- `VMLX_SSM_DISK_CACHE_NAMESPACE`

I've updated my local start scripts accordingly.

Closing this PR. Thanks — the upstream version answers all 4 of the open questions I raised in the PR description.

PR #26, opened 2026-03-24
feat: Prompt Lookup Decoding — Phase 2/3 speculative decode

Fail-Safe:
## Summary

Implements Prompt Lookup Decoding (PLD) speculative acceleration for vMLX with **adaptive auto-tune** that makes `--enable-pld` safe on all models. PLD is net-positive on slow models (+4–7% on Qwen3.5-27B-Claude-4.6-6bit) and auto-disables within 1 token on fast models.

- **Phase 2**: batched speculative verification at T=0
- **Phase 3**: probabilistic acceptance at all temperatures — no output corruption
- **Phase 5**: d0 pre-check + K=2 + bfloat16 fix + adaptive auto-tune + scalar logprob + O(1) ngram index
- **Opt-in via `--enable-pld`**: defaults to off; auto-tune handles the rest

## Phase 5 optimizations

Initial benchmarks showed PLD was **-22% to -36% on all models**. Root cause: SSM sequential processing penalty (verify costs ~4× on hybrid models) + per-cycle remove/insert overhead (~20ms).

Nine changes brought PLD from **-22% to +4–7%** on Qwen3.5-27B-Claude-4.6-6bit and safe across all 5 tested models:

| # | Change | Effect |
|---|--------|--------|
| 1 | K=2 for hybrid models | Verify cost ~4× → ~1.75×. Amortizes ~20ms overhead over 4 tokens. |
| 2 | d0 pre-check | `forward_logprobs[d0] > -2.0` before `remove()`. Filters ~60% of cycles. |
| 3 | bfloat16 numpy fix | Cache rollback crashed on bf16. Eliminated ~40% emergency re-inserts. |
| 4 | Raw logprob lookup | Replaced full-vocab softmax with single array index for d0 check. |
| 5 | Backward n-gram scan | Most recent prior match → better contextual relevance. |
| 6 | Adaptive auto-tune | TCP slow-start window (1→2→4→...→487). Congestion/no-opportunity → auto-disable. Probes after 5×interval. |
| 7 | Scalar logprob acceptance | Phase 3 acceptance uses `logsumexp` + `math.exp` instead of materializing full 150K softmax. Lazy d1..dK evaluation. |
| 8 | O(1) n-gram hash index | `NgramIndex` replaces O(n) backward scan with hash lookup. 140× faster on 12K-token sequences. Incrementally updated per token. |
[...]

PR #149, opened 2026-05-07
feat(pld): hybrid partial-accept replay for SSM models (#134)

### LINKED ISSUE #134: Perf: PLD verify-cost on hybrid SSM models — proposal for SSM checkpoint/replay (extends #26)
## Context

PR #26 shipped PLD Phase 2/3 with a documented headroom of **+4–7%** on Qwen3.5-27B-Claude-4.6-6bit (hybrid SSM/ATT) and **~0% to −2%** on the other hybrid Qwen3.5/3.6 27B variants. The PR description identifies the root cause:

> Verify costs ~4× on hybrid models … K=2 for hybrid models | Verify cost ~4× → ~1.75×

That ~4× verify cost is the SSM sequential-processing penalty: GatedDeltaNet layers can't batch K+1 verify tokens like the 16 full-attention layers do. Once K=2 is forced, the speculation budget per cycle is too small to amortize the per-cycle ~20 ms remove/insert overhead even with d0 pre-check + adaptive auto-tune.

## Proposal

Replace the verify-pass-then-rollback approach for the 48 SSM layers with **snapshot-before-draft + replay-on-accept**:

1. Before drafting, snapshot the per-layer `ArraysCache` state via the existing deep-copy primitive at `vmlx_engine/utils/ssm_companion_cache.py:103-372` (already does `mx.array(a) * 1` + `mx.eval()` per-layer for the SSM companion cache).
2. Run the draft probe / verify on the **16 full-attention layers only** — i.e. skip SSM for the verify pass; the draft probe doesn't need correct SSM output to score the accept/reject decision because the accept gate uses the lm_head logits which are conditioned on the joint state at the snapshot point. This is the trick: do a *sliced* verify.
3. On accept, replay the accepted draft tokens through SSM layers incrementally from the snapshot. On reject, restore the SSM snapshot.

The verify cost on the SSM half collapses from ~4× per token to amortized 1× per accepted token (snapshot is ~50 MB / 48 layers × ~1 MB; restore is free; replay is the same cost as the original AR baseline).

This is the same trick DFlash uses on CUDA via `ggml_gated_delta_net_tree` (see #76 / Luce-Org/lucebox-hub `dflash/`). Filing as a separate issue because PLD's chain-verify case is materially simpler than DFlash's tree verify and ships first.

## Expected impact

On hybrid 27B thinking-mode workloads (e.g. Qwen3.6-27B-Heretic2 9.4bit): push PLD from +4–7% to **+15–25%** by removing the SSM verify-cost ceiling. Combined with the existing K=2 + d0 + adaptive auto-tune, this should make PLD net-positive on every hybrid 27B variant tested in #26 (currently ~0% / −1% / −2%).

## Files

- Snapshot primitive: `vmlx_engine/utils/ssm_companion_cache.py:103-372`
- Verify pass: PLD path inside `vmlx_engine/scheduler.py` (post-#26)
- Hybrid layer detection: existing `_is_hybrid` in `scheduler.py:307`

## Risk

Low–medium. The "skip SSM in verify pass" trick is mathematically sound when accept gate uses target-model lm_head logits; the SSM state only becomes load-bearing for *future* tokens, which the replay-on-accept covers. Correctness gate: greedy T=0 byte-equality vs PLD-off.

## References

- #26 description (Phase 5 optimization journey table)
- DFlash paper (Luce-Org port arXiv 2602.06036)
- `notes/prompt-lookup-decoding.md`

st-adam:
## Summary

- Fixes the token loss on hybrid SSM/ATT models when `0 < num_accept < K` in PLD partial-reject path
- Adds `Scheduler._replay_ssm_forward()` to restore caches to N, replay accepted tokens, advance to N+K', emit K'+1 tokens instead of 1
- Default ON; opt-out: `VMLX_DISABLE_PLD_REPLAY=1`
- New `/health` field `pld_ssm_replay.{enabled,attempts,emitted,failures}`
- 6 unit tests in `tests/test_pld_ssm_replay.py`

## Problem

PR #26 PLD on hybrid models (48 GatedDeltaNet + 16 full-attention): with K=2, a partial accept (`num_accept=1`) still emits only 1 correction token because SSM state cannot be trimmed — both caches must rewind to N.

## Solution

After partial rejection, restore to N, replay `drafts[:num_accept]` forward through the full model. Both caches reach N+num_accept. Emit `drafts[:num_accept] + [bonus_token]` — same as the full-accept path, minus the extra K-K' tokens.

## Expected gain

+5-10% on top of PR #26's +4-7% on hybrid models. Full PLD target for hybrid moves from +4-7% toward the +15-25% cited in #134.

## Test plan

- [ ] `pytest tests/test_pld_ssm_replay.py -v` — 6 unit tests pass
- [ ] `pytest tests/test_ssm_companion_cache.py -v` — existing tests unaffected
- [ ] Live model: `VMLX_DISABLE_PLD_REPLAY=1` vs unset — byte-equal at T=0, higher tok/s unset

Fixes #134


Ari4ka (approved):

st-adam:
Rebased onto current `main` (1.5.32 series, base commit `9cfbeb24`) per @jjang-ai's note on #134 ("PRs in most recent stable versions"). No code changes — clean rebase over the 50 intervening commits.

Verified post-rebase:
- `tests/test_pld_ssm_replay.py` — 6/6 passing
- `vmlx_engine.scheduler` imports cleanly; `Scheduler._replay_ssm_forward` present
- Upstream's `memory_limits` refactor (v1.5.31) integrated cleanly into our scheduler.py touch points — no shadowing of `get_metal_ws_guard_threshold` / `get_effective_metal_working_set_bytes`

PR is `CLEAN` / `MERGEABLE` against the 1.5.32 base.

st-adam:
Rebased onto v1.5.44 (upstream/main as of 2026-05-20). Resolved conflicts with hybrid TQ KV path commit (c61b6e46). Tests: 6/6 pass.

jjang-ai:
This is the strongest issue-fix candidate in the current PR batch. Thank you for keeping it rebased and tying it to #134.

I am not merging it directly into the active release-hardening tree yet because that tree is dirty with DSV4/cache/tool/settings work, and this touches `vmlx_engine/scheduler.py` plus `/health` telemetry. The right next step is an isolated worktree validation:
- re-check the current scheduler/cache interaction against hybrid SSM/TQ paths;
- run the PR unit tests plus adjacent scheduler/cache tests;
- run a live hybrid SSM PLD proof showing partial-accept replay counters and no output regression;
- only then pull it into the release lane.

Keeping this marked as a candidate fix rather than a feature-only PR.

jjang-ai:
I did one more concrete review pass because this is the strongest issue-fix candidate in the open PR batch.

One thing that needs fixing before merge: `tests/test_pld_ssm_replay.py` copies a local `_replay_ssm_forward()` implementation instead of importing/exercising `Scheduler._replay_ssm_forward`. That means the tests can pass while the production scheduler method diverges or is wired incorrectly.

Please retarget those tests so they execute the actual production method, or factor the replay helper into a small importable function that both scheduler and tests call. The current fake-cache approach is fine, but the assertion needs to cover the real code path that will ship.

After that, I still want the isolated validation mentioned above: adjacent scheduler/cache tests plus a live hybrid SSM PLD proof showing replay attempts/emitted counters and no output regression.

jjang-ai:
Reviewed for the current release-hardening pass. I am not merging/adapting this into the immediate release branch yet.\n\nThe problem statement is credible: hybrid SSM/attention PLD partial accept can lose accepted draft tokens when cache rollback has to rewind to N. Credit to @st-adam for isolating that design and proposing replay.\n\nReason held for this release:\n- this changes scheduler cache replay behavior in the same hot area as current prefix/paged/L2/hybrid-SSM release risks;\n- the submitted tests include standalone/copied replay logic rather than proving the production Scheduler path end-to-end;\n- no current live proof here shows a hybrid SSM model exercising 0 < num_accept < K replay with healthy follow-up generation and unchanged cache telemetry;\n- /health pld_ssm_replay would need source/API contract coverage before shipping.\n\nCurrent-source guard verification I ran instead: existing PLD/spec guard slice passed (7 passed, 581 deselected), including the PLD non-MLLM short-circuit, current speculative continuous-batching warnings, DSV4 runtime speculative suppression, and prompt-lookup docs matching current scheduler integration.\n\nLeaving open as future feature work; not safe to close as merged/adapted.

st-adam:
**Addressing review feedback — production code path now tested directly.**

The test-local copy of `_replay_ssm_forward` has been removed. The fix:

- New `vmlx_engine/utils/pld_replay.py` — canonical `replay_ssm_forward()` with minimal deps (lazy `mlx_lm` imports, `contextlib.nullcontext` fallback if `generation_stream` unavailable in test env)
- `Scheduler._replay_ssm_forward` is now a 2-line delegation wrapper calling `replay_ssm_forward` from the utils module
- `tests/test_pld_ssm_replay.py` imports `from vmlx_engine.utils.pld_replay import replay_ssm_forward` — tests exercise the real production code path

If the production method diverges, the import itself will surface it (wrong signature → TypeError; wrong module path → ImportError).

**Tests: 6/6 pass** (same assertions, now over real code).

---

**Re: isolated worktree validation + live PLD proof**

The live hybrid SSM proof (showing `pld_ssm_replay.attempts/emitted` counters) requires a server started with a hybrid SSM model (e.g. Qwen3.6-27B). That cannot be automated here — it needs the target hardware and model weights. `tests/benchmark/test_pld_acceptance.py` includes the `partial_accept_stress` task (Task 5) specifically for this; the check is:

```bash
python tests/benchmark/test_pld_acceptance.py --port 8080
curl -s http://127.0.0.1:8080/health | python3 -m json.tool | grep -A5 pld_ssm_replay
```

Expected after Task 5: `pld_ssm_replay.attempts > 0`, `emitted > attempts` (each replay emits K'+1 tokens, not 1). Happy to coordinate on hardware access if that helps move the validation forward.

Ari4ka (approved):

st-adam:
**Follow-up: special-token filter added.**

Latent issue identified during PR #150 review work: `prompt_lookup.find_draft_tokens()` and `NgramIndex.find_drafts()` had no special-token filtering. n-gram lookup over the prompt could propose tokens like pad/image-pad/vision-start as drafts. Verify usually rejects, but truncating at the first excluded ID is cheaper and removes the risk entirely.

New commit on this branch:
- `prompt_lookup._truncate_at_excluded()` — helper
- `find_draft_tokens(exclude_token_ids=None)` — backward-compatible param
- `NgramIndex.find_drafts(exclude_token_ids=None)` — same
- `Scheduler._build_pld_excluded_token_ids()` — collects IDs from tokenizer/processor (`pad_token_id`, `image_token_id`, `image_pad_id`, `vision_start_token_id`, `vision_end_token_id`, `additional_special_tokens_ids`)
- Wired through both PLD call sites (scheduler.py:5193 and :6626)
- EOS/BOS intentionally NOT excluded — legitimate end-of-decode signals already gated by verify

**Tests:** 20/20 pass (`tests/test_prompt_lookup_filter.py` 14 new + `test_pld_ssm_replay.py` 6 existing).

Ari4ka (approved):

PR #155, opened 2026-05-08
fix: JANG model compat — MiniMax sanitize + MoEGate quantize bypass

pperezrubio:
## Summary

- **MiniMax sanitize crash fix**: JANGTQ CRACK stores expert weights as `tq_bits/tq_norms/tq_packed`. `jang_loader` dequants experts 0–2 (producing `.weight` keys), but `Model.sanitize` only checks expert 0 then list-comprehensions all 256 experts — crashing at expert 3+. Patch uses safe `.pop(key, None)` instead of `.pop(key)`.
- **MoEGate quantize bypass**: When loading JANG-quantized MoE models (e.g. Nemotron 120B), `nn.quantize` crashes on `MoEGate` layers that have no quantizable weights. Patch skips quantization for `MoEGate` modules during weight loading.

Both patches are in `vmlx_engine/runtime_patches/jang_model_compat.py` and auto-install via the existing `__init__.py` patch mechanism. They follow the same pattern as existing runtime patches in the repo.

## Testing

- **MiniMax M2.7 JANGTQ (57GB)**: Loads in 19s via JANGTQ fast path, benchmarks 28–40 tok/s on M4 Max 128GB. Intelligence 10/15, coding 6/7.
- **Nemotron 3 Super 120B JANG_2L (43GB)**: 8/8 tool calling tests pass, 29–34 tok/s.
- **Ling 2.6 Flash JANGTQ2 (29GB)**: Text/coding work. Tool calling fails (2-bit repetition loop — known JANGTQ2 limitation, not related to these patches).

## Files Changed

- `vmlx_engine/runtime_patches/jang_model_compat.py` — new file with two patches
- `vmlx_engine/runtime_patches/__init__.py` — auto-install entry updated

### DIFF
--- a/vmlx_engine/runtime_patches/__init__.py
+++ b/vmlx_engine/runtime_patches/__init__.py
@@ -15,12 +15,16 @@ never corrupts the bundle.
 Registered patches (auto-installed on ``import vmlx_engine.runtime_patches``):
   * ``kimi_k25_mla``          — Kimi K2.6 fp32 MLA L==1 SDPA cast
   * ``deepseek_v4_register``  — DSV4 mlx_lm.models.deepseek_v4 registration
+  * ``jang_model_compat``     — JANG-quantized model fixes (MiniMax sanitize,
+                                MoEGate quantize)
 """
 
 # Eagerly install every patch on first import so they land before any
 # model load path walks config.json.model_type. Safe: each installer guards
 # against already-patched state internally.
 from . import deepseek_v4_register  # noqa: F401
+from . import jang_model_compat as _jang_compat
 from . import kimi_k25_mla as _kimi_k25_mla
 
+_jang_compat.install()
 _kimi_k25_mla.install()
--- a/vmlx_engine/runtime_patches/jang_model_compat.py
+++ b/vmlx_engine/runtime_patches/jang_model_compat.py
@@ -1,0 +1,266 @@
+# SPDX-License-Identifier: Apache-2.0
+"""JANG-quantized model compatibility patches.
+
+Two runtime patches that fix crashes when loading dealignai JANG-quantized
+models through ``mlx_lm``:
+
+A) **MiniMax sanitize fix** — ``mlx_lm.models.minimax.Model.sanitize``
+   crashes with ``KeyError`` on JANGTQ CRACK models where expert weights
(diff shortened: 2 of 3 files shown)


jjang-ai:
Thank you for digging into the JANG MiniMax / MoEGate compatibility class. I reviewed this against the current release-hardening source.

I do not think this PR is mergeable as-is now:
- it conflicts with current source;
- the MoEGate/Nemotron path is already handled in current `vmlx_engine/utils/tokenizer.py`, `vmlx_engine/utils/jang_loader.py`, and covered by tests around stale MoEGate quantization;
- the global `nn.quantize` monkeypatch is too broad for release-hardening because it can change unrelated model-family loading behavior.

I am keeping credit for the issue class in the PR audit. If you still have a current reproducible MiniMax sanitize failure on latest `main`, please rebase and narrow the patch to the current loader path with a model/config fixture test instead of a global quantize monkeypatch.

jjang-ai:
Reviewed this against the current release-hardening branch. The two failure classes you identified are real and are now covered in source, but we adapted them without taking the global runtime monkeypatch shape.\n\nAccepted/covered with credit to @pperezrubio:\n- packed JANGTQ bundles route through the native TurboQuant loader before fallback sanitize paths;\n- JANG quantization predicates skip modules without to_quantized, so MoEGate-like modules are not handed to nn.quantize;\n- Nemotron-H / Nemotron-H-v2 stale MoEGate quantization entries are sanitized at config-load time;\n- MiniMax uses the registered minimax_m2 parser/reasoning route, with tests ensuring we do not hide MiniMax loops behind a forced family repetition-penalty floor.\n\nNot accepted from this PR:\n- auto-installing a global runtime_patches/jang_model_compat.py monkeypatch;\n- deleting the hybrid SSM generic L2 disk-cache store guard. That path still needs a separate hybrid SSM/block-disk equivalence proof before changing current cache policy.\n\nVerification run today:\n- tests/cross_matrix/run_jang_model_compat_contract.py -> status=pass\n- tests/test_jang_model_compat_pr155.py + TestNemotronLatentMoePatch + MiniMax no-hidden-repetition-floor contract -> 12 passed\n\nClosing because the accepted compatibility findings have been adapted into current source with regression coverage. Thanks for the report and patch direction.

PR #162, opened 2026-05-12
perf: low-risk hot-path cleanups (PR-A of 3)

st-adam:
## Summary

Tier-1 (low-risk) pure-implementation perf cleanups. No behaviour change — sampling outputs, cache contents, and request contracts are unchanged.

Base: `9cfbeb24` (current `main`, 1.5.32 series).

This is **PR-A of 3** in a planned perf-audit series. PR-B will land env-gated per-chunk allocator-clear and boundary precompute changes; PR-C will land the sampler buffer reuse + async-eval pipelining.

## What's in this PR

| # | finding | fix |
|---|---------|-----|
| 1 | `scheduler.py:6527-6531` PLD uses `uid not in ab.uids` + `ab.uids.index(uid)` — two O(N) scans per speculative step. | New `MLLMBatch.index_of(uid)` / `has_uid(uid)` backed by a lazily-built dict, invalidated on `filter` / `extend`. PLD path now O(1). |
| 2 | `mllm_batch_generator.py:_make_request_sampler` does `from mlx_lm.sample_utils import make_sampler` and `make_logits_processors` per new request. | Hoisted to module-level `from mlx_lm.sample_utils import make_logits_processors, make_sampler`. |
| 3 | `utils/ssm_companion_cache.py:_clone_states` does `c.lengths = mx.array(c.lengths) * 1` — the `* 1` is a force-materialize trick that emits a needless Metal op since the next line already calls `_mx_materialize(c.lengths)`. | Dropped the multiply. Kept the `mx.array(...)` constructor and the explicit `_mx_materialize` call. |
| 4 | `utils/ssm_companion_disk_store.py` post-`safetensors.load_file` per-state `deepcopy`. Safetensors already returns fresh arrays, and the L1 fetch path (`SSMCompanionCache.fetch`) always wraps the disk result through `_clone_states` before handing it to the model. The disk-store deepcopy was a second copy doubling RAM on every L2 hit. | Dropped the deepcopy + the `from copy import deepcopy` import. Added a comment pointing at the upstream `_clone_states` callsite. |

## What's deliberately not in this PR

- `paged_cache.py:72` token-hash format change. Investigated — it would invalidate persisted L2 block-disk-store hashes (`block_disk_store.py` keys on `BlockHash`), forcing a one-time cache cold-start for users with `--enable-block-disk-cache`. Realistic gain on the hash itself was <0.1% of TTFT (128 hash calls × ~10µs at 32K with 256-tok blocks). Not worth the disk-cache churn.
[...]

PR #163, opened 2026-05-12
perf: env-gated prefill allocator clear + chunked-loop tightening (PR-B of 3)

st-adam:
## Summary

Tier-2 perf cleanups on the long-context chunked-prefill path. All behaviour-preserving; the headline change (skipping per-chunk `mx.clear_cache()`) is **opt-in** so existing users see zero change.

Base: `9cfbeb24` (current `main`, 1.5.32 series). Independent of [PR-A #162](https://github.com/jjang-ai/vmlx/pull/162) (low-risk hot-path cleanups).

This is **PR-B of 3** in the perf-audit series. PR-C (sampler buffer reuse + async pipelining) follows.

## Changes in `MLLMBatchGenerator._process_prompts` chunked-text branch

| # | finding | fix |
|---|---------|-----|
| 7 | Per-chunk `mx.clear_cache()` at line 2722 flushes the Metal allocator free-list between every chunk of a long-context prefill. Redundant given the v1.5.31 98 % working-set guard. | Env-gated. New `--prefill-keep-alloc` CLI flag (or `VMLX_PREFILL_KEEP_ALLOC=1`) skips the call. **Default behaviour unchanged.** |
| 8 | Per-chunk `next((b for b in ssm_boundaries if processed < b <= processed + chunk_size and b not in captured), None)` is an O(B) generator scan. For a 32K prompt with `prefill_step_size=1024` and 4 boundaries that's 32 generator allocations + 128 conditional checks per request. | Sorted-list + pointer index; pointer advances monotonically as the loop progresses. O(1) per chunk. |
| 9 | `getattr(request, '_original_token_ids', None) or input_ids[0].tolist()` evaluated on every boundary hit; the fallback triggers a Metal→CPU sync. | Computed once before the loop when `_original_token_ids` is unset. `_maybe_capture_clean_ssm_boundary` accepts a Python list, so the contract is unchanged. |
| 10 | `mx.eval([c.state for c in cache if hasattr(c, 'state')])` per chunk — `hasattr` reflection over the full cache layer list every iteration. | Precompute `_state_layers = [c for c in cache if hasattr(c, "state")]` once before the loop; reuse inside. |

## Out of scope (kept unconditional, justified inline)

- The image-prefill `mx.clear_cache()` at the one-shot VLM entry (line 2849 in upstream) is a deliberate pre-guard against the Metal command-buffer OOM on large pixel inputs (server-kill scenario). Not redundant with the 98 % guard for that codepath.
- Error-recovery `mx.clear_cache()` calls in the OOM-recovery and SSM re-derive branches are low-frequency and off the hot path. Left untouched.

## Tests

- `tests/test_perf_prefill_loop.py` — 8 new unit tests:
  - `test_chunk_loop_uses_sorted_boundary_pointer` — boundary pointer + sorted list present
  - `test_chunk_loop_precomputes_state_layers` — `_state_layers` reused per chunk
  - `test_chunk_loop_hoists_all_tokens_tolist` — `_hoisted_all_tokens` present
  - `test_chunk_loop_env_gates_clear_cache` — `if not _prefill_keep_alloc: mx.clear_cache()` gate present
  - `test_cli_flag_propagates_to_env` — `--prefill-keep-alloc` sets `VMLX_PREFILL_KEEP_ALLOC=1`
  - `test_prefill_keep_alloc_env_off_by_default` — default behaviour unchanged
  - `test_prefill_keep_alloc_env_recognised` — truthy spellings work (`1`, `true`, `yes`, `on`, case-insensitive)
  - `test_boundary_pointer_advances_past_captured` — synthetic pointer-advance check
- Adjacent regression sweep: `tests/test_memory_limits.py` 8/8, `tests/test_ssm_companion_cache.py` 14/14, `tests/test_batching.py` 54/56 (same 2 pre-existing pytest-asyncio infra failures, unaffected).

## Expected gain

With `--prefill-keep-alloc`:
- **5-12 % TTFT improvement** on long prompts (>8K tokens), depending on chunk count. The allocator-clear elision is the bulk; boundary precompute + tolist hoist + state-layer precompute are constant-factor.

Without the flag:
- Same behaviour as today. Boundary pointer + state-layer precompute + tolist hoist still apply (these are unconditional optimizations — they reduce Python overhead without changing the Metal command stream).

## Test plan

- [x] `pytest tests/test_perf_prefill_loop.py -v` — 8/8 pass
- [x] `vmlx serve --help` shows `--prefill-keep-alloc`
- [x] CLI flag → env var propagation verified in unit test
- [ ] Live model: TTFT at 8K / 16K / 32K with `VMLX_PREFILL_KEEP_ALLOC` on vs off; verify no OOM
- [ ] `bench/live_speed_probe.py` before/after on the standard long-context prompt

### DIFF
--- a/tests/test_perf_prefill_loop.py
+++ b/tests/test_perf_prefill_loop.py
@@ -1,0 +1,130 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Unit tests for PR-B prefill-loop perf cleanups.
+
+Algorithm-only: no real model load, no real cache layers.  We test the
+helper logic in isolation and assert the source-level invariants
+(env-gate present, boundary pointer used, hoisted tolist) via inspection.
+
+Run:
+    .venv/bin/python -m pytest tests/test_perf_prefill_loop.py -v
+"""
+
+from __future__ import annotations
+
+import inspect
+import os
+
+import pytest
+
+
+def _src() -> str:
+    import vmlx_engine.mllm_batch_generator as mod
+
+    return inspect.getsource(mod)
+
+
+def test_chunk_loop_uses_sorted_boundary_pointer():
+    """The chunk loop must walk a sorted-boundary pointer, not a per-chunk
+    `next(generator)` linear scan over the boundary list."""
+    src = _src()
+    # Generator-style lookup is gone.
+    assert "for b in ssm_boundaries" not in src or (
+        "_sorted_boundaries" in src
+    ), "stale generator-style next(...) lookup still present"
+    # Pointer-style names are present.
+    assert "_sorted_boundaries" in src, "boundary pointer list missing"
+    assert "_boundary_idx" in src, "boundary pointer index missing"
+
+
+def test_chunk_loop_precomputes_state_layers():
+    """The per-chunk `mx.eval([c.state for c in cache if hasattr…])` must be
+    replaced by a precomputed `_state_layers` list."""
+    src = _src()
+    assert "_state_layers" in src, "_state_layers precompute missing"
+    assert "mx.eval([c.state for c in _state_layers])" in src, (
+        "per-chunk eval over precomputed _state_layers missing"
+    )
+
+
+def test_chunk_loop_hoists_all_tokens_tolist():
+    """The `getattr(request, '_original_token_ids', None) or input_ids[0]
+    .tolist()` Metal→CPU sync must be hoisted out of the chunk loop."""
+    src = _src()
+    assert "_hoisted_all_tokens" in src, "hoisted tolist variable missing"
+
+
+def test_chunk_loop_env_gates_clear_cache():
+    """`mx.clear_cache()` inside the chunk loop must be gated on
+    `_prefill_keep_alloc`, the env-var-driven flag."""
+    src = _src()
+    assert "VMLX_PREFILL_KEEP_ALLOC" in src, "env var name missing"
+    assert "_prefill_keep_alloc" in src, "gate variable missing"
+    assert "if not _prefill_keep_alloc:" in src, "gate wrapper missing"
+
+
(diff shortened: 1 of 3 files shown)


jjang-ai:
Thank you for the prefill allocator/chunk-loop perf work. I am deferring this from the current release-hardening lane rather than merging it now.

Reason: it is performance work on the long-context prefill path, and the active tree is currently focused on correctness gates for DSV4/tools/cache/settings. Before merge this needs current-source tests plus a live long-context TTFT/memory-pressure comparison showing the opt-in allocator behavior does not regress large-model stability.

jjang-ai:
Accepted the safe slice on current main. Credit to @st-adam for the prefill allocator/chunk-loop cleanup work.\n\nMerged as adapted commit:\n- 5a5cd2c2 perf: add opt-in prefill loop cleanup\n\nWhat landed:\n- sorted SSM boundary lookup in the chunked text prefill loop\n- precomputed state-layer eval list\n- hoisted token-list materialization for SSM boundary capture\n- opt-in --prefill-keep-alloc / VMLX_PREFILL_KEEP_ALLOC=1 allocator tuning flag; default behavior stays unchanged\n- no-heavy regression contracts in tests/test_perf_prefill_loop.py\n\nFresh verification before close:\n- tests/test_perf_prefill_loop.py: 7 passed\n- CLI UDS/socket binding slice: 3 passed\n- SSM companion + memory-limit tests: 59 passed\n- selected batching prefill/cache tests: 12 passed\n- combined selected gate: 88 passed, 43 deselected\n- py_compile for changed Python files: pass\n- git diff --check: pass\n\nI did not pull in PR #162/#164/#149/#150 here; those remain separate because they touch scheduler/sampler/speculative/cache behavior and need their own proof.

PR #164, opened 2026-05-12
perf: sampler shared-params fast-path + single_batch async pipelining (PR-C of 3)

st-adam:
## Summary

Tier-3 perf cleanups on the per-step sampling and single-batch decode loop. No behaviour change — sampling outputs and cache contents are unchanged.

Base: `9cfbeb24` (current `main`, 1.5.32 series). Independent of [PR-A #162](https://github.com/jjang-ai/vmlx/pull/162) and [PR-B #163](https://github.com/jjang-ai/vmlx/pull/163).

This is **PR-C of 3** in the perf-audit series.

## Changes

| # | finding | fix |
|---|---------|-----|
| 12 | `MLLMBatchGenerator._step` per-request sampler loop: for each request, slice `logits[i:i+1]`, call `self._make_request_sampler(req)(...)`, then `mx.concatenate(tokens, axis=0)`. Cost grows with batch size even when every request shares identical sampling params. | New `_batch_shares_sampler_params(requests)` helper. When every request has the same `(temperature, top_p, top_k, min_p)` AND none has a non-trivial repetition penalty, dispatch the batched `self.sampler(logits)` in one shot. The per-request path remains the fallback. |
| 14 | `single_batch_generator._clone_array` uses `value + mx.zeros_like(value)` — two Metal ops (zero-init + add) when one would do. | Replaced with `1 * value`. Single Metal op, same independent-buffer semantics. |
| 13 | `single_batch_generator` decode step issues two sequential `mx.eval(...)` calls for `current_token` and `current_logprobs`. | Replaced with one `mx.async_eval(current_token, current_logprobs)`. The implicit sync still happens on the immediate `_token_to_int(current_token)` call below, so observable behaviour is unchanged — but the Metal scheduler can overlap both materialisations. |

## Out of scope (deferred to a focused follow-up PR)

- Sampler closure's `mx.array(_prompt_list + _req.output_tokens)` per call — the original plan proposed a pre-allocated MLX int32 buffer on the request with amortised-doubling growth. That work has a wider blast radius than the other items here and reads better as its own PR (touches the sampler closure, the request lifecycle, and possibly the response-token bookkeeping).

## Tests

- `tests/test_perf_sampler_async.py` — 8 new unit tests:
  - `test_shared_params_fast_path_all_equal` — fast-path active when all match
  - `test_shared_params_fast_path_one_differs` — disabled on any mismatch
  - `test_shared_params_fast_path_repetition_penalty_disables` — rep penalty disables (per-request token history)
  - `test_shared_params_fast_path_first_request_repetition_penalty` — short-circuits on first req
  - `test_shared_params_fast_path_empty_returns_false` — empty batch
  - `test_shared_params_fast_path_repetition_penalty_none_is_ok` — penalty=None treated as 1.0
  - `test_single_batch_clone_array_uses_single_mul` — `1 * value` form present, `value + mx.zeros_like(value)` form absent
  - `test_single_batch_uses_async_eval_for_token_pair` — `mx.async_eval(current_token, current_logprobs)` present, stale `mx.eval(current_logprobs)` absent
- Adjacent regression: `tests/test_memory_limits.py` 8/8.

## Expected gain

- **3-6 %** steady-state decode improvement on multi-seq batches where all requests share sampling params (the typical OpenAI-compat / dev-loop case).
- **2-4 %** decode improvement on SimpleEngine (async pipelining).
- One Metal op + one zero-tensor alloc dropped per per-token snapshot in the single-batch path.

## Test plan

- [x] `pytest tests/test_perf_sampler_async.py -v` — 8/8 pass
- [x] Helper is importable at module scope; type-correct API
- [ ] Live model: 4× concurrent requests with shared params produces byte-identical tokens compared to the per-request fallback at T=0
- [ ] `bench/live_speed_probe.py --num-requests 4` before/after on the standard prompt

### DIFF
--- a/tests/test_perf_sampler_async.py
+++ b/tests/test_perf_sampler_async.py
@@ -1,0 +1,113 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Unit tests for PR-C sampler + async pipelining perf cleanups.
+
+Covers:
+- `_batch_shares_sampler_params` fast-path predicate (shared / mixed /
+  repetition-penalty / empty).
+- `single_batch_generator._clone_array` uses `1 * value` (no `zeros_like`).
+- `single_batch_generator` decode step uses `mx.async_eval` for the
+  token + logprobs pair.
+
+Run:
+    .venv/bin/python -m pytest tests/test_perf_sampler_async.py -v
+"""
+
+from __future__ import annotations
+
+import inspect
+import types
+
+import pytest
+
+
+def _make_req(temperature=0.0, top_p=1.0, top_k=0, min_p=0.0, repetition_penalty=1.0):
+    """Build a duck-typed request that satisfies the helper's getattr probes."""
+    req = types.SimpleNamespace()
+    req.temperature = temperature
+    req.top_p = top_p
+    req.top_k = top_k
+    req.min_p = min_p
+    req.repetition_penalty = repetition_penalty
+    return req
+
+
+def test_shared_params_fast_path_all_equal():
+    from vmlx_engine.mllm_batch_generator import _batch_shares_sampler_params
+
+    reqs = [_make_req() for _ in range(4)]
+    assert _batch_shares_sampler_params(reqs) is True
+
+
+def test_shared_params_fast_path_one_differs():
+    from vmlx_engine.mllm_batch_generator import _batch_shares_sampler_params
+
+    reqs = [_make_req() for _ in range(3)]
+    reqs.append(_make_req(temperature=0.7))
+    assert _batch_shares_sampler_params(reqs) is False
+
+
+def test_shared_params_fast_path_repetition_penalty_disables():
+    from vmlx_engine.mllm_batch_generator import _batch_shares_sampler_params
+
+    # All same params, but one has a non-trivial rep penalty.
+    reqs = [_make_req() for _ in range(3)]
+    reqs.append(_make_req(repetition_penalty=1.2))
+    assert _batch_shares_sampler_params(reqs) is False
+
+
+def test_shared_params_fast_path_first_request_repetition_penalty():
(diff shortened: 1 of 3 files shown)


jjang-ai:
Thanks for the sampler/async pipelining work. I am deferring PR-C from the current release-hardening lane.

It currently conflicts with `main`, and this path touches per-step sampling behavior. Before merge it needs a rebase, focused sampler tests on current source, and live output-equivalence/perf evidence for representative model families.

jjang-ai:
Adapted the safe subset in main as b3669837 (perf: adapt safe sampler fast path). Credit to @st-adam for PR #164 and the sampler/clone cleanup direction.\n\nWhat landed:\n- shared sampler-params predicate for active MLLM batches;\n- corrected fast path uses the request-scoped sampler from batch.requests[0], not self.sampler(logits), so shared non-default request sampler params are preserved;\n- SingleBatchGenerator._clone_array now uses 1 * value instead of value + mx.zeros_like(value);\n- no-heavy regression coverage in tests/test_perf_sampler_async.py.\n\nWhat I did not land from the PR:\n- the async_eval current_token/current_logprobs hunk. Current main has _eval_on_stream() stream rehoming in single_batch_generator, so copying the older async hunk would bypass stream-safety behavior without a separate audit.\n\nVerification before close:\n- red first tests/test_perf_sampler_async.py: 7 failed / 1 passed before source changes; final 8 passed;\n- selected scheduler/memory/batching gate: 83 passed, 8 deselected;\n- single-batch/cache focused gate: 23 passed;\n- py_compile changed files: pass;\n- git diff --check: pass.\n\nNo live model speed proof was run for this subset, so this is credited as a safe no-heavy adaptation, not a live perf claim.

PR #168, opened 2026-05-18
feat(cli): add --uds flag for Unix domain socket serving

efortin:
**Summary**

Adds a --uds <path> flag to vmlx-server serve that binds the server to a Unix domain socket instead of TCP. This is what embedded sidecars need: each Alfred install gets an isolated server without having to hunt for a free TCP port.

**Implementation**

Forwards the path straight to uvicorn's uds= kwarg through the existing serve runner.
TCP host/port path is unchanged when --uds is omitted — fully backward compatible.

**Usage**

`vmlx-server serve --uds /tmp/alfred-vmlx.sock`

### DIFF
--- a/vmlx_engine/cli.py
+++ b/vmlx_engine/cli.py
@@ -1214,8 +1214,12 @@ def serve_command(args):
         print()
 
     # Start server
-    print(f"Starting server at http://{args.host}:{args.port}")
-    uvicorn.run(app, host=args.host, port=args.port, log_level=log_level.lower())
+    if args.uds:
+        print(f"Starting server on Unix domain socket {args.uds}")
+        uvicorn.run(app, uds=args.uds, log_level=log_level.lower())
+    else:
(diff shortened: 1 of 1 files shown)


jjang-ai:
Thanks for the Unix domain socket support PR. This is useful integration work, but I am deferring it from the current release-hardening lane because it is a feature surface rather than a blocker fix.

Before merge I would want a small test/docs pass covering CLI parsing, `uvicorn.run(..., uds=...)`, how the panel/API clients should display or consume a UDS endpoint, and whether TCP host/port health checks should be skipped or translated when UDS mode is active.

jjang-ai:
Closing this because Unix-domain-socket serving has now been incorporated into the current release-hardening tree with attribution rather than direct-merging this older branch. Current source has vmlx-engine serve --uds, _uvicorn_bind_kwargs coverage in tests/test_cli_commands.py, and docs/changelog entries. Thank you @efortin; the changelog credits PR #168.

PR #173, opened 2026-05-23
fix(spec): sequential K+1 verify forwards — root cause + workaround for MLLM PLD logit drift

st-adam:
## Summary

**Root-caused** the MLLM PLD output drift reported in PR #171 via empirical diagnostic. Multi-token forward on `mlx-community/SmolVLM-Instruct-bf16` produces DIFFERENT logits than K+1 sequential single-token forwards on the same starting cache, even though cache writes match between paths.

### Diagnostic (committed: `tests/benchmark/diagnose_multi_token_forward.py`)

Measured on smolvlm pure-attention model:
```
pos 0: max_diff=0.250  argmax_M=11089  argmax_S=11089  (FP off, top-1 matches)
pos 1: max_diff=0.266  argmax_M=7526   argmax_S=7526   (FP off, top-1 matches)
pos 2: max_diff=0.375  argmax_M=30     argmax_S=36     (argmax FLIPS)
cache_keys_last_position max_diff = 0.000000  (cache writes byte-equal)
```

This is the root cause of all observed MLLM PLD divergence — including the `"TheTheThe"`, `"return        return"` patterns. The bug is in mlx_lm/mlx_vlm's multi-token `__call__` (likely future-position leakage in attention mask construction), not in PLD's accept/reject logic.

### Fix

Replace the single multi-token verify forward in `MLLMBatchGenerator._step_speculative` with K+1 sequential single-token forwards. Each call uses input shape (B, 1) — identical to standalone `_step`'s call path, so byte-equivalent by construction.

Cost: K+1× model forwards per spec round. PLD net-gain holds when acceptance > 1/(K+1) (e.g., > 33% for K=2). Auto-disable (`VMLX_PLD_MIN_ACCEPTANCE=0.30` default) protects throughput on bad-fit workloads.

### Test plan

- [x] `pytest tests/ --tb=no -q` — 95/95 unit tests pass (mock model updated to support sequential verify)
- [x] Diagnostic script run on smolvlm: confirms multi-token path is the root cause
- [ ] Live byte-equality on a DETERMINISTIC pure-attention MLLM model — smolvlm itself shows T=0 non-determinism in baseline (3 same-prompt runs → 3 different outputs), so byte-equality measurement on this model is unreliable independent of PLD

### Known limitation

Smolvlm baseline non-determinism prevents direct PLD-on vs PLD-off byte-equality validation. The diagnostic + sequential-verify fix is theoretically sound (matches standalone decode path byte-for-byte at the model-call level). Live validation against a deterministic test model is the open follow-up.

PLD on MLLM remains **default-OFF** (opt-in via `VMLX_ENABLE_MLLM_PLD=1`) until a deterministic baseline model is identified for byte-equality validation.

### Stacks on

PRs #149, #150, #171.

### DIFF
--- a/notes/prompt-lookup-decoding.md
+++ b/notes/prompt-lookup-decoding.md
@@ -550,3 +550,65 @@ gate has been removed; PLD now fires at all temperatures.
    batch_size > 1. This is unmeasured and should be verified before relying
    on PLD in high-concurrency deployments. The warning is documented in the
    `_try_speculative_decode` docstring.
+
+---
+
+## Hybrid partial-accept replay (vmlx#134)
+
+### Problem
+
+On hybrid SSM/ATT models (e.g. 48 GatedDeltaNet + 16 full-attention layers) case (b)
+of the rollback logic previously wasted accepted drafts. When `0 < num_accept < K`
+(partial reject), the code restored both caches to position N and emitted only 1
+correction token — discarding the `num_accept` tokens the model had already validated.
+
+The root cause: `ArraysCache` state cannot be arbitrarily trimmed, so both caches must
+rewind to a consistent checkpoint. Before this fix that checkpoint was always N (zero,
+pre-verify), even when some drafts were accepted.
+
+### Solution
+
+`Scheduler._replay_ssm_forward()` is a new staticmethod that:
+
+1. Restores all `ArraysCache` layers to the saved pre-verify snapshot (position N).
+2. Rewinds all trimmable `KVCache` layers to `pre_verify_offset` (N) via numpy roundtrip.
+3. Runs one model forward pass on `accepted_tokens` (shape `[1, num_accept]`), advancing
+   both caches to `N + num_accept` as a side-effect.
+4. Returns `True` on success, `False` on any exception (with best-effort cache restore).
+
+When replay succeeds, case (b1) emits `drafts[:num_accept] + [bonus_token]` — the same
+output as the full-accept path but for a shorter accepted prefix.
+
+When replay fails or is disabled, case (b2) falls back to the original correction-only
+path (1 token emitted).
+
+### Expected gain
+
+On hybrid models with K=2 and a 50% partial-accept rate, the replay path turns
+single-token correction rounds into two-token rounds, roughly doubling effective
+throughput on those rounds. Combined with PR #26's baseline PLD gain (+4–7%), the
+expected net improvement is **+5–10%** on top of existing PLD on hybrid models.
+
+### Configuration
+
+| Control | Default | Effect |
+|---------|---------|--------|
+| `VMLX_DISABLE_PLD_REPLAY=1` | OFF (replay enabled) | Disables replay, falls back to b2 path |
+| `config.pld_replay_enabled` | `True` | Python-level toggle (overrides env var) |
+
+### Telemetry
+
+`/health` now includes a `pld_ssm_replay` field when the scheduler has replay counters:
+
+```json
+{
+  "pld_ssm_replay": {
+    "enabled": true,
+    "attempts": 142,
+    "emitted": 421,
+    "failures": 0
+  }
+}
+```
+
+`[PLD:3b1f]` summary lines include `replay=emitted/attempts fail=N` when replay is active.
--- a/notes/prompt-lookup-decoding.md
+++ b/notes/prompt-lookup-decoding.md
@@ -612,3 +612,137 @@ expected net improvement is **+5–10%** on top of existing PLD on hybrid models
 ```
 
 `[PLD:3b1f]` summary lines include `replay=emitted/attempts fail=N` when replay is active.
+
+---
+
+## MLLM Batched PLD (PRs #149/#150 — issue #134/#135)
+
+### Background
+
+The simple-engine PLD path (above) operates only on requests served via the
+non-batched `Scheduler` class. The user's typical workload (`vmlx serve ...
+--continuous-batching` with an MLLM-detected model) uses `MLLMScheduler` +
+`MLLMBatchGenerator`, which previously had **zero PLD code**. PR #150
+ports PLD into that batched MLLM path. PR #149's `replay_ssm_forward`
+helper is reused for hybrid SSM rollback.
+
+### Dispatch path
+
+`MLLMBatchGenerator._next()` (vmlx_engine/mllm_batch_generator.py) routes
+each step to one of three paths:
+
+| Condition | Path | When |
+|-----------|------|------|
+| `VMLX_ENABLE_BATCHED_SPEC=1` + `--speculative-model` set | `_step_speculative` with draft model | Draft spec configured |
+| `config.pld_enabled` AND model is non-hybrid (or hybrid opt-in) | `_step_speculative` with PLD candidate source | `--enable-pld` set |
+| Otherwise | `_step` (standard 1-token forward) | Default |
+
+`MLLMScheduler.__init__` calls `batch_generator.configure_pld_spec(enabled, excluded_token_ids)`
+at startup. The excluded-token set (image-pad, vision markers, pad) is built from
+tokenizer/processor attributes — same V0.6 filter as PR #149's simple-engine path.
+
+### In-batch verify
+
+`_step_speculative(batch, K)` runs **one** target-model forward on input shape
+(B, K+1) where each row is `[seed_i, d0_i, ..., d_{K-1}_i]`. Per-row drafts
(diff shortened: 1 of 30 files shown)


st-adam:
**Phase C + Layer 1/2/3/5 tests landed.**

### Phase C — TurboQuant short-circuit

Detect TurboQuant KV cache at the entry to `_step_speculative` and fall back to `_step`. Prevents the post-completion `extract()` crash observed when `--kv-cache-quantization q4` + `--enable-pld` + MLLM. Logs once at INFO. Workaround until TQ-compatible per-row writeback is implemented.

### Layer 1 — Logit-equivalence tests (slow, real model)

`tests/test_mllm_pld_logits_equivalence.py` — 3 tests on `mlx-community/Llama-3.2-1B-Instruct-4bit`:
- `test_sequential_matches_multi_token_via_argmax` — argmax matches at all positions
- `test_sequential_matches_multi_token_via_logits_tolerance` — logits within 1e-2 FP tolerance
- `test_sequential_matches_multi_token_cache_state` — cache offset + last-position keys match

**All 3 PASS on Llama-3.2.** Proves mlx_lm's multi-token forward IS mathematically correct. The smolvlm-specific divergence (PR #172 Phase B diagnostic) is therefore in **mlx_vlm's wrapper layer**, not mlx_lm core. Sequential verify is the right workaround for VLM-routed paths.

### Layers 2/3/5 — TQ + invariants + edge cases (mock, fast)

`tests/test_mllm_pld_tq_and_invariants.py` — 11 tests:
- TQ cache short-circuit (skip + log-once)
- Cooldown invariants (decrement, no underflow)
- Offset monotonicity under sequential verify
- Telemetry counter monotonicity
- Edge cases: K=0, B=0, prefill phase

### Test totals

| Layer | Tests | Status |
|---|---|---|
| 1 — Logit equivalence (slow) | 3 | PASS on Llama-3.2 |
| 2 — Unit fix tests | covered in step_speculative + tq_and_invariants | PASS |
| 3 — Property invariants | covered in tq_and_invariants | PASS |
| 5 — Edge cases | covered in tq_and_invariants | PASS |

**Combined: 117 tests passing** (95 from prior PRs + 11 new TQ/invariant + 8 new B>1 from PR #171 + 3 slow logit-equivalence).

### PR #149 live-path discovery

While attempting Layer 6 live regression on Llama, discovered that PR #149's `Scheduler._try_speculative_decode` is **not reachable via `vmlx serve`** — both batched and non-batched modes route through `AsyncEngineCore` / `BatchedEngine` paths that bypass it. `pld_ssm_replay.attempts` stays at 0 in all tested CLI configurations. PR #149's PLD was validated via mock unit tests; live activation requires direct API use with `AsyncEngineCore(..., scheduler=Scheduler(...))`. This needs a follow-up to either:
  1. Wire PR #149's PLD into the AsyncEngineCore's Scheduler instance, OR
  2. Mark PR #149's code path as library-internal (used only by direct API)

### Not yet done

- Layer 4 / Layer 6 LIVE byte-equality on a deterministic MLLM — smolvlm is non-deterministic at T=0; need to identify a deterministic VLM (Qwen2-VL or Phi-3-vision are candidates, ~4 GB download)
- Phase D default-on flip — blocked on Layer 4 PASS

st-adam:
**Layer 4 live byte-equality findings: sequential verify alone is NOT enough.**

Tested on `mlx-community/Qwen2-VL-2B-Instruct-4bit` (deterministic at T=0 — 3/3 same-prompt runs byte-identical, confirmed baseline).

**Result: 1/4 PASS, 3/4 FAIL** (same pattern as smolvlm).

```
--- short_factual ---  PASS (2 chars)
--- repetitive ---    FAIL  'AAA BBB CCC.' vs 'AAA BBB.'
--- code ---          FAIL  missing newline + token duplications ("1 1\n1")
--- json ---          FAIL  heavy token duplication in JSON structure
```

Telemetry: spec.batched.steps=20, tokens_emitted=39, acceptance_rate=0.215. EMA dropped below threshold → cooldown triggered after warmup.

**Interpretation**:
- mlx_lm multi-token forward on Llama-3.2 produces byte-identical logits to sequential (Layer 1 logit-equivalence: 3/3 PASS). So multi-token forward is correct in mlx_lm core.
- Sequential verify in MLLMBatchGenerator's `_step_speculative` produces correct logits per call (each is a standalone `_step` call shape (B,1) — byte-identical to standalone decode by construction).
- BUT MLLM PLD byte-equality still fails. Bug isn't in the multi-token forward OR the sequential verify. Something else in the PLD pipeline corrupts state.

**Suspected remaining bugs** (need follow-up):
- Per-request `_pld_ngram_index` aliasing across requests in same session
- Cache state corruption between PLD-on and the standard `_step` fallback (auto-disable cooldown transition)
- `response.extra_tokens` ordering / detokenizer state leak between steps
- VLM model wrapper layer differences (Qwen2-VL wrapper does extra processing)

**Conclusion for this PR**:
- Sequential verify is **theoretically correct** (Layer 1 tests prove logit equivalence on Llama)
- It is **not sufficient** to achieve T=0 byte-equality in MLLM batched path
- MLLM PLD remains **default-OFF** via `VMLX_ENABLE_MLLM_PLD=1` opt-in
- TurboQuant short-circuit prevents the q4-related crash
- Diagnostic + unit-test infrastructure shipped for follow-up investigation

**Test count: 117 passing** (95 mock + 11 new TQ/invariant + 8 B>1 + 3 slow logit-equivalence).

**Next session needs**:
1. Bisect the MLLM PLD pipeline by disabling individual components (auto-disable, n-gram lookup, extras emission, response builder) one at a time
2. Identify the exact step where output diverges from PLD-off
3. Either fix the corrupting component OR re-scope PLD on MLLM to a structurally simpler approach (e.g., draft-model spec only, no PLD)

st-adam:
## Seed Staleness Fix — Root Cause of Byte-Equality Failure

### Finding

Bisected the MLLM PLD pipeline and found the **primary** root cause of the 3/4 byte-equality failures:

**`_step_speculative` used `req.last_token` as the verify forward seed, but `req.last_token` is ONLY updated inside `_step_speculative` itself.**

When PLD falls back to standard `_step` (no n-gram drafts found, cooldown transition, or prompt processing), `batch.y` gets updated by `_step` but `req.last_token` stays at the value from the last successful spec step. When PLD re-engages on the next step, it feeds the **wrong token** to the model as the seed, corrupting the KV cache and causing all subsequent output to diverge.

### Why 3/4 prompts failed

N-gram matching is intermittent — early in generation, few output tokens → frequent fallback to `_step`. After fallback, the next PLD step used the stale seed. Only highly repetitive text (where every step finds n-gram matches) avoided the bug.

### Fix

```python
# Before (bug):
seed_int = int(req.last_token) if req.last_token is not None else int(batch.y[i].item())

# After (fix):
seed_int = int(batch.y[i].item())
```

`batch.y` is maintained by BOTH `_step()` and `_step_speculative()`, so it's always correct regardless of which path ran on the previous step.

### Tests

3 new regression tests covering:
- `test_seed_uses_batch_y_not_last_token` — stale vs correct seed
- `test_seed_after_cooldown_transition` — PLD re-engagement after cooldown
- `test_seed_when_last_token_none` — first PLD step

All 109 PLD tests pass.

### Status

Combined with the sequential K+1 verify forwards (previous commit), this addresses both:
1. Multi-token forward logit divergence (VLM-specific)
2. Seed staleness causing cache corruption after fallback transitions

Next: live byte-equality validation on a deterministic VLM to confirm 4/4 PASS.

Ari4ka (approved):

st-adam:
Closing as **superseded by prod's native-MTP MLLM batched-spec path** (`mllm_batch_generator.py` `MLLMNativeMTPState`/`_native_mtp_*` + scheduler `_try_speculative_decode`).

This PR built batched/MLLM PLD speculative decoding *inside* `MLLMBatchGenerator`; prod independently re-solved that capability via embedded native-MTP. Every core symbol this PR introduces (`VMLX_ENABLE_BATCHED_SPEC`, `_step_speculative`, `configure_pld_spec`, `_snapshot_ssm_per_row`, `should_use_speculative_batched`, etc.) now returns 0 hits in upstream/main, and the files it rewrites diverged 700–3000 lines each — so this is architectural replacement, not a textual conflict, and a rebase would be meaningless.

Salvageable orthogonal bits, if still wanted, will be cut fresh from `main` as small standalone PRs rather than rebased from here: the `prompt_lookup` special-token n-gram exclusion filter (note: also carried by #149) and `paged_cache.release_last_K_blocks_from_seq`.

PR #184, opened 2026-06-02
docs: add Homebrew cask install instructions

SSakutaro:
Adds Homebrew cask install instructions for the macOS desktop app.
The existing DMG and source build instructions are left unchanged.
https://github.com/Homebrew/homebrew-cask/pull/267095/

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,12 @@ The vMLX inference server is now running at `http://0.0.0.0:8000` with an OpenAI
 
 Get [MLX Studio](https://github.com/jjang-ai/mlxstudio/releases/latest) -- a native macOS app with chat UI, model management, image generation, and developer tools. No terminal required. Just download the DMG and drag to Applications.
 
+Or install it with Homebrew:
+
+```bash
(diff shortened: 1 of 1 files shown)


fettpl:
@jjang-ai Homebrew has an accepted `vmlx` cask, and the current maintenance path should be a normal cask bump once the public artifact is clean.

I checked the current `v1.6.3` Sequoia DMG for the Homebrew update path today. The SHA256 matches release notes and the DMG contains `vMLX.app`, but the app currently fails both `codesign` and Gatekeeper assessment:

```text
vMLX.app: invalid signature (code or signature have been modified)
In architecture: arm64
```

```text
vMLX.app: internal error in Code Signing subsystem
```

I opened the distribution blocker here: https://github.com/jjang-ai/mlxstudio/issues/125

Once that is fixed, the Homebrew PR should be a version/SHA-only update to the existing `Homebrew/homebrew-cask` `vmlx` cask, preserving the current `jjang-ai/vmlx` release URL.

PR #210, opened 2026-06-17
fix(flash-moe): support Step-3.7-Flash (step3p7) expert offload

st-adam:
## Problem

`--flash-moe` (and `--smelt`) cannot offload experts for **Step-3.7-Flash** (`step3p7` VLM, text backbone `step3p5`) JANG bundles. On `dealignai/Step-3.7-Flash-JANG_2L-CRACK` (201B MoE, 81.6 GB), `--flash-moe` logged `ExpertIndex built: 0 MoE layers` and fell back to a full load.

Three distinct gaps, each fixed:

1. **ExpertIndex regex** (`smelt_loader._EXPERT_KEY_RE`) matched only `switch_mlp` keys. step3p7 raw safetensors store routed experts as `…language_model.layers.N.moe.{gate,up,down}_proj` — the `switch_mlp` name only appears *after* `jang_loader`'s load-time remap, but `ExpertIndex` scans raw keys → 0 experts. Added a 5th alternative for the bare `.moe` parent. Router `.moe.gate.weight` and the always-resident `.share_expert.*` correctly do **not** match.

2. **Routing detection** (`_detect_routing_style`) didn't know `Step3p5MoE`, whose gate (`Step3p5MoEGate`) returns `(topk_indices, topk_weights)` directly (sigmoid + router bias + scaling + norm inside). It defaulted to `"softmax"` → `mx.softmax()` invoked on a tuple → `RuntimeError`. Mapped `Step3p5MoE`/`Step3p7MoE` → `"pre_routed"`.

3. **Shared expert** (`FlashMoEBlock.__call__`) looked up only `shared_expert(s)`; step3p5 names it `share_expert` (singular, combined as a plain add `y = routed + share_expert(x)`). The always-on shared branch was dropped from all 42 MoE layers → incoherent output. Now checks the `share_expert(s)` spellings too.

## Validation

On `dealignai/Step-3.7-Flash-JANG_2L-CRACK` (M3 Max 128 GB, vmlx 1.5.61):

- Before: `ExpertIndex built: 0 MoE layers` → no offload.
- After: `ExpertIndex built: 42 MoE layers, 288 experts/layer, expert=77.29GB backbone=10.33GB`; `Flash MoE enabled`.
- Coherent output (`7+5 → "12"`; `"capital of France … 13×4=52"`).
- Resident: ~10 GB (slot-bank 64) … ~61 GB (slot-bank 8192). Survives 33k-token prompts with memory flat — the prompt size that OOM-crashed full residency at a 112 GB wired limit.

## Notes / scope

- Routing fix relies on `Step3p5MoEGate` returning `(indices, weights)`; the `pre_routed` path already handles that interface.
- The latent `mx.softmax(..., precise=True)` kwarg in the `softmax` routing branch is **not** touched here (step3p7 no longer hits it); flagging for a separate fix if other families take that path on newer mlx.
- `--smelt` shares the same `ExpertIndex`/routing detection and benefits from (1)+(2); its partial-expert/TQ path is unchanged.


st-adam:
Added a second commit: **bf16 expert scales/biases bit-reinterpret fix** (`flash_moe_loader._pread_tensor`).

Separate root cause from the step3p7 loader gaps, but surfaced while validating flash-moe on `dawncr0w/Step-3.7-Flash-oQ3-MLX`: `_NP_DTYPE_MAP["BF16"] = np.float16` and the loader never reinterpreted the bytes, so bf16 scales/biases were corrupted ~88× → garbage expert output. Bundles with float16 scales (e.g. JANG) were unaffected (F16→np.float16 is a no-op). Fix bitcasts the float16-read words to real bfloat16 at both return sites. Verified bit-exact (max_abs_diff 0.0) vs the true bf16 tensor on oQ3 layer-20 expert 0.

This is a **general flash-moe correctness fix** independent of step3p7 — any bf16-scale MoE bundle benefits.

Note: oQ3 still can't serve end-to-end on the MLLM path due to a *separate* flash-moe×async-prefill stream issue — filed as a standalone issue.

PR #214, opened 2026-06-21
Fix MiniMax-M3 VL API routing

hornsan1:
## Summary
- avoid MiniMax-M3 VL request-time AutoProcessor imports by using the native vMLX image preprocessor plus tokenizer image-token expansion
- align the MiniMax-M3 vision tower with the official 3D RoPE contract and full vision attention path
- quantize MiniMax-M3 VL JANG vision/projector modules by mapping runtime vision.* paths back to bundle-side names
- keep M3 VL image prompts in a single prefill forward so pixel_values and image-token positions co-occur

## Validation
- py_compile on touched runtime and test files
- direct MiniMax-M3 VL checks for native preprocessing/token expansion, quant path mapping, single-forward VL prefill, and 3D RoPE shapes
- local server smoke tested text and image requests back-to-back on MiniMax-M3-REAP22-Coder via vMLX

### DIFF
--- a/tests/test_minimax_m3_cache_paths.py
+++ b/tests/test_minimax_m3_cache_paths.py
@@ -615,6 +615,47 @@ def test_single_batch_m3_chunks_long_prompt_before_final_sample():
     assert model.calls == [[1, 2, 3], [4, 5, 6], [7, 8]]
 
 
+def test_single_batch_m3_vl_prefills_image_prompt_in_one_forward():
+    from vmlx_engine.utils.single_batch_generator import SingleBatchGenerator
+
+    class _FakeVLModel:
+        def __init__(self):
+            self.calls = []
+
+        def make_cache(self):
+            return []
+
+        def __call__(self, tokens, cache=None, pixel_values=None, image_grid_thw=None):
+            self.calls.append(
+                (
+                    tokens.tolist()[0],
+                    pixel_values is not None,
+                    image_grid_thw is not None,
+                )
+            )
+            return mx.zeros((1, tokens.shape[1], 8), dtype=mx.float32)
+
+    model = _FakeVLModel()
+    gen = SingleBatchGenerator(
+        model,
+        max_tokens=1,
+        sampler=lambda _logits: mx.array([3], dtype=mx.int32),
+        prefill_step_size=3,
+        stream=None,
+    )
+
+    gen.insert(
+        [[1, 2, 3, 4, 5, 6, 7, 8]],
+        pixel_values=[mx.zeros((1, 1), dtype=mx.bfloat16)],
+        image_grid_thw=[mx.array([[1, 1, 1]], dtype=mx.int32)],
+    )
+    prompt_responses, generation_responses = gen.next()
+
+    assert len(prompt_responses) == 1
+    assert generation_responses == []
+    assert model.calls == [([1, 2, 3, 4, 5, 6, 7, 8], True, True)]
+
+
 def test_scheduler_uses_minimax_m3_logits_sampler_for_msa_cache(monkeypatch):
     from vmlx_engine.models.minimax_m3.cache import MiniMaxM3SparseCache
     from vmlx_engine.request import SamplingParams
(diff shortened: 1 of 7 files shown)


jjang-ai (commented):
Reviewed against an independent fix I landed for the same root cause (3D-RoPE) on `feat/minimax-m3-runtime` (commit 73e92e0ef). Core change here is correct and matches the official contract — two things need behavioral verification before merge, and the native-preprocessor switch is the highest-risk part.

**3D-RoPE (`minimax_m3_vl.py`) — correct ✅**
This is the actual root cause. The vision tower uses 3-axis (T,H,W) RoPE per `transformers.models.minimax_m3_vl.MiniMaxM3VL3DRotaryEmbedding`. The old code was a 2-axis (H,W) rope with the wrong `axis_dim` that rotated all 80 dims, so the H/W frequencies landed in the wrong head-dim slots — per-patch color survived but spatial structure scrambled (red circle → "red arrow/blocks", hallucinated grids). This PR's version (3-axis; `axis_dim = 2*((2*(head_dim//2)//3)//2)`; merge-block-ordered `hi`/`wi`; `concat([freqs,freqs])`; pass-through tail past `3*axis_dim`) matches the upstream contract exactly — identical to what I implemented and verified.

I verified this exact RoPE end-to-end on **REAP22-Coder and Coder-Small (both 2-bit)**: red circle → "A red circle"; a structured scene reads correctly as "red circle on the left, blue square on the right, text below" — shapes, colors, left/right/below positions, and OCR all correct; robust across sizes 112→896, aspect ratios, PNG/JPEG/WebP, JPEG q5 (heavy), RGBA/grayscale, multi-turn and multi-image. Direction confirmed; the 2-bit quant was never the cause.

**Native preprocessor (`m3_vl_preprocess` → `preprocess_image`) — needs numeric-parity check ⚠️ (highest risk)**
Dropping the request-time HF `AutoProcessor` for native `preprocess_image` is a good perf/robustness win, but if native `pixel_values`/`grid_thw` don't match HF *exactly* (smart-resize rounding, patch/merge order, channel layout, mean/std normalization) VL degrades even with the rope fixed. Please add a parity test asserting `preprocess_image(img)` ≈ `AutoProcessor(img)` (pixel_values allclose + identical `grid_thw`) over a few sizes/aspect ratios — not just shape checks.

**Quant path mapping (`jang_loader`) — good catch ✅**
Mapping runtime `vision.*` → bundle `vision_tower.*` / `multi_modal_projector.*` / `patch_merge_mlp.*` so `nn.quantize` sees the vision linears is correct. Worth a one-line assert that the vision linears are actually `QuantizedLinear` post-load.

**Validation gap**
"3D RoPE shapes" + "smoke tested text and image" confirms it runs, not that it's correct. Recommend the behavioral checks above (actual correct shape/position/OCR across sizes) as the acceptance bar.

**Overlap heads-up**
`feat/minimax-m3-runtime` already carries the same 3D-RoPE fix (73e92e0ef) plus gemma4 VL / MM3 parser / Qwen3.6 MXFP4-MTP fixes for v1.5.68 — these two will conflict in `minimax_m3_vl.py`; let's reconcile so we don't double-apply or diverge.

Net: core fix is correct and confirmed; good to merge once native-preprocessor parity is verified.

hornsan1:
Addressed the verification feedback in follow-up commit 4f2f564b.

Added:
- `tests/test_minimax_m3_vl_processor_parity.py`: opt-in native `preprocess_image` vs HF `AutoImageProcessor` parity over multiple sizes/aspect ratios, asserting identical `image_grid_thw` and pixel `assert_allclose`. It is guarded behind `VMLINUX_RUN_M3_VL_PROCESSOR_PARITY=1` + `VMLINUX_M3_VL_PARITY_MODEL` because the HF processor imports torch.
- `test_minimax_m3_vlm_quant_candidates_select_vision_linears`: constructs a tiny `vision.multi_modal_projector.linear_1`, runs MLX `nn.quantize`, and asserts it becomes `nn.QuantizedLinear` through the MiniMax-M3 VL path mapping.

Local validation:
- `py_compile` passed for touched runtime/test files.
- Direct quant selection check passed: `vision.multi_modal_projector.linear_1` becomes `nn.QuantizedLinear`.
- Direct 3D-RoPE shape check passed: grid `[1,12,18]` produces `(216,78)` cos/sin and preserves q/k `(1,216,16,80)`.
- Direct MiniMax-M3 VL preprocessing/prefill function checks passed.

Caveat: I could not complete the live HF parity run on this machine because `import torch` alone hung for >2 minutes and had to be interrupted. That is also the production reason for keeping the HF processor out of the request-time path. The parity test is now present as the explicit maintainer acceptance gate on an environment where the HF torch image processor imports normally.

hornsan1:
Retried the native preprocessor parity check and found the reviewer concern was valid.

Result before the latest fix:
- `image_grid_thw` matched for all tested sizes/aspect ratios.
- `pixel_values` shapes matched.
- Pixel allclose failed after resized inputs because the native path used PIL bicubic while the bundled MiniMax processor uses torch bicubic interpolation with `align_corners=False, antialias=True`.
- Observed max abs deltas ranged roughly `0.0074` to `0.2118` on resized cases.

Fix in commit `0e3c3764`:
- `preprocess_image` now uses `torch.nn.functional.interpolate(..., mode="bicubic", align_corners=False, antialias=True)` for resize, matching `MiniMaxM3VLImageProcessor.resize` without importing HF `AutoProcessor` or torchvision.
- The parity test now compares native output against a direct torch reference implementation of the bundled processor math, avoiding the local `AutoImageProcessor -> torchvision -> torch._dynamo` import trap.

Validation after fix:
- Torch-reference parity passes at `rtol=1e-5, atol=1e-5` for sizes `(112,112)`, `(113,251)`, `(640,360)`, `(896,512)`, `(47,999)`.
- Max abs diff after fix: <= `4.77e-07`; grids and shapes identical.
- `py_compile`, RoPE shape check, and quantized vision-linear selection check still pass.

PR #218, opened 2026-06-25
engine: fix MLA paged prefix-cache reconstruct for GLM-5.2

Deviad:
Two bugs in BlockAwarePrefixCache blocked the MLA paged prefix cache for GLM-5.2 (every warm turn re-prefilled the entire prefix):

1. _extract_block_tensor_slice (CacheList positional sub-cache branch): when a positional sub-cache had seq_len==0 for a block (legitimate MLA case where a layer's secondary head group accumulated 0 tokens for the prefix), the store emitted ('skip',). Reconstruct then saw the sub as all-skip across all blocks and aborted the WHOLE rebuild, forcing a full re-prefill. Fix: store an empty zero-seq slice [..., start:0, :] (preserves feature dim) as 'kv'/'quantized_kv' instead of ('skip',). The empty array is non-None so the downstream live-cache validator accepts it, and the standard 'kv' rebuild produces a real empty KVCache matching the live state.

2. reconstruct_cache (CacheList KV sub-rebuild padding): MLA's compressed latent has asymmetric k/v feature dims (e.g. k last-dim 64, v 512). The padding derived pad_shape from ck.shape and reused it for v_pad, crashing mx.concatenate when offset % step != 0. Fix: build pad shapes per-tensor (k_pad_shape from ck.shape, v_pad_shape from cv.shape).

Verified on GLM-5.2 JANGTQ_K (53K BLUE-FALCON prompt, --use-paged-cache): cold 1404.7s -> warm 59.6s (24x speedup, decode-only), BLUE-FALCON-48217 recovered on both turns. Both fixes opt-in via --use-paged-cache; default memory-aware cache runs are byte-identical to upstream.


jjang-ai (commented):
Grounded this against main before reviewing — verdicts differ per hunk.

**Hunk 3 (`reconstruct_cache` per-tensor pad shapes) — correct, should land.** The asymmetry is real in our tree: GLM-5.1's `make_cache` is `[CacheList(KVCache(), KVCache()) for _ in self.layers]` (`panel/scripts/patches/deepseek_v32.patched.py:659`), sub 0 stores `kv_latent`/`k_pe` = 512/64-wide (`:202`), and sub 1 (the DSA indexer) stores 128-wide keys with a **zero-width values tensor** `mx.zeros([b, 1, s, 0])` (`:100`) — so keys/values last-dims differ on *both* subs (your writeup has the dims swapped: keys end in 512, values in 64, but the asymmetry claim holds). With `KVCache.step = 256` and 64-token blocks, `offset % step != 0` on 3 of every 4 block boundaries — this is a hot path, not an edge case. One requirement:

- Add a regression test: CacheList sub with keys last-dim ≠ values last-dim (cover 512/64 *and* the 128/0 indexer sub), non-256-aligned offset, assert `reconstruct_cache` doesn't raise and produces correctly-shaped `.keys`/`.values`. This padding path has zero test coverage before or after this PR (`grep -r pad_shape tests/` → nothing).

**Hunks 1–2 (skip → empty slice) — premise unproven, scope too broad as written.**

1. Model naming: "GLM-5.2" is not registered anywhere in this repo — `model_configs.py:538-551` registers `glm5` for `model_type glm_moe_dsa`, and it's GLM-5.1. Which model/config actually reproduced this? If it was an unregistered model falling through generic detection, that changes what "verified" means here.
2. The claimed trigger ("a layer's secondary head group accumulated 0 tokens") doesn't obviously exist for glm5: both CacheList subs update in lockstep every forward — `cache[0]` at `deepseek_v32.patched.py:202`, `cache[1]` at `:100` (the indexer updates its keys *before* the `<= index_topk` early-exit at `:101-102`). Please add a unit repro showing which sub emitted `("skip",)` on your 53K run and why; without it these two hunks are speculative hardening attached to a real fix (hunk 3).
3. The guard removal is **unconditional** for every CacheList positional sub, not scoped to your zero-seq case. Before: `start_idx >= actual_end` (sub genuinely shorter than the requested block range) → reconstruct aborts → safe full re-prefill. After: empty slices get stored and reconstruct rebuilds a sub whose token count undercounts the block range — **silent truncation where the old code failed safe**. Scope the empty-slice conversion to `seq_len == 0` and keep the abort for partially-short subs, or demonstrate the under-fill can't happen.
4. The `quantized_kv` hunk (`prefix_cache.py:2521-2534`) is dead code in default config: MLA models auto-disable KV-cache quantization (`scheduler.py` `_detect_mla` gate, ~691-745, mirrored in the MLLM scheduler; override is `VMLX_ALLOW_MLA_KV_QUANT=1`). Was that override actually exercised, or is this hunk untested?
5. The fix doesn't survive the disk tier: `block_disk_store.py:873-898` (`cache_list` write loop) only handles `skip|kv|cumulative` sub-tags — no `quantized_kv` branch — so a quantized empty sub silently degrades back to skip after an SSD round-trip. Separately, please verify a zero-size tensor (empty seq slice, or the `[b,1,S,0]` indexer values) round-trips through the safetensors write in that lane at all, and add a store → disk → reconstruct regression test for the empty-sub case.
6. For the record: glm5's `cache_type` is `"kv"` (non-structural), so paged defaults OFF and this entire path is opt-in via `--use-paged-cache` — the PR correctly notes defaults are byte-identical, just making the blast radius explicit for other reviewers.

Happy to take hunk 3 + its regression test as a standalone PR if you'd rather split — that part is mergeable on its own merits.

PR #223, opened 2026-07-01
Fix Qwen hybrid prompt-cache reuse with native MTP

jakimoai:
## Summary

Fixes Qwen/Ornith hybrid SSM warm prompt-cache reuse when native MTP is active. The fix keeps reconstructed attention KV in native mutable cache form for the native-MTP request instead of immediately recompressing that restored prompt-cache state into TurboQuant.

Live validation on Qwen3.6 JANG artifacts eliminated repeated `!` failures on warm `paged+ssm` hits while preserving prefix cache, SSM companion state, block-disk cache, TurboQuant storage/cold/live paths, and native MTP.

## Problem

Qwen3.6 27B JANG artifacts could produce a coherent first response, then fail on subsequent same-prompt warm-cache runs by emitting long repeated `!` runs. The failure reproduced even with deterministic sampling and occurred on real warm `paged+ssm` prompt-cache hits with native MTP active.

Controls narrowed the issue:

- cache bypass avoided the symptom, but did not satisfy the goal of correct warm-cache reuse
- clean prompt-boundary cache storage was necessary, but not sufficient
- disabling native MTP cleared the warm-cache failure
- disabling TurboQuant KV cleared the native-MTP warm-cache failure
- guard-off ablation brought the failure back even with the clean-store, validator, and native-MTP position-id fixes retained

## Root Cause

PROVEN for the tested Qwen3.6 JANG VLM/native-MTP surface: reconstructed hybrid SSM prompt-cache attention KV was being recompressed into `TurboQuantKVCache` immediately before native MTP verify/replay/rollback.

That produced the wrong cache representation contract for native MTP. Native MTP expects prompt-boundary cache layers it mutates, snapshots, restores, trims, and replays against to behave like ordinary mutable KV cache layers. On a warm hybrid SSM hit, the SSM companion state is already path-dependent and restored for the prompt boundary; recompressing only the reconstructed attention side into the split TurboQuant live representation before speculative MTP mutation made the state subtly corrupt.

## Fix

- Store generic hybrid SSM paged prompt-cache records from clean prompt-boundary prefill instead of truncating post-generation live cache state.
- Accept valid single-sequence `TurboQuantKVCache` prompt-cache records where `offset` is set and `_idx` is still unset.
- Pass absolute `position_ids` through native MTP backbone seed, verify, and replay calls when the language model supports them.
- For hybrid SSM prompt-cache hits that will enter native MTP, keep reconstructed attention KV layers native for that request instead of recompressing them to TurboQuant before MTP verify/replay/rollback.
- Add focused regressions for clean hybrid SSM store, single-sequence TurboQuant validation, native-MTP absolute position ids, and the native-KV warm-hit guard.

## Why This Shape

This is deliberately narrow. TurboQuant is still enabled generally; the PR only avoids one unsafe transition: recompressing reconstructed hybrid prompt-cache KV into TurboQuant immediately before native MTP runs.

The fix does not disable:

- prefix cache
- paged `paged+ssm` reuse
- SSM companion state
- block-disk cache
- storage-boundary TurboQuant
- cold/live TurboQuant paths outside this reconstructed warm-hit/native-MTP boundary
- native MTP

## Live Proof

| Interface / Harness | Model / Config | Result |
|---|---|---|
| OpenAI Chat Completions streaming repro | Qwen3.6 27B JANG_4M, MLLM path, paged+SSM, TurboQuant, native MTP D3 | Patched cold/default, warm/default, and warm/temp0 passed with real warm `paged+ssm` hits and native MTP active |
| Cross-quant contrast | Qwen3.6 27B JANG_4K | Baseline failed warm/default and warm/temp0; patched passed both warm hits |
| Fresh quant contrast | Fresh current-converter Qwen3.6 27B JANG_4M | Baseline failed warm/default and warm/temp0; patched passed both warm hits |
| Non-MLLM text-path contrast | Qwen3.6 27B MLX 4-bit text path | Baseline and patched both passed, supporting that the bug is not generic to every hybrid/TurboQuant cache path |
| Isolated guard ablation benchmark | JANG_4K, original JANG_4M, fresh JANG_4M | Patched: 0 repeated-`!` marker runs out of 48 warm `paged+ssm` hits. Guard-off ablation: 41/48 failures |

Evidence captured:

- spawn args: yes
- startup logs: yes
- `/health`: yes
- streaming chunks/reasoning/content: yes
- native MTP telemetry: yes
- cache-hit telemetry: yes
- memory/peak-memory readings: yes

## Performance Notes

The expanded patched-vs-guard-off benchmark showed a consistent TTFT improvement: patched TTFT was lower in every benchmark cell.

Total t/s was generally equal or better for patched, but it is less decisive than TTFT because many ablation runs generated corrupted repeated-`!` output and are not quality-equivalent throughput comparisons. The safe conclusion is:

- TTFT improved consistently on the tested warm-cache/native-MTP path.
- Total t/s did not show a clear regression and usually improved, but exact t/s attribution is partially confounded by corrupted ablation outputs.

## Regression / Adjacent Coverage

- Existing JANG_4K and original JANG_4M artifacts covered non-fresh artifact behavior.
- Fresh JANG_4M covered current-converter artifact behavior.
- MLX 4-bit text path covered a non-MLLM contrast and confirmed warm hybrid/TurboQuant cache reuse is not globally broken.
- Prose, code, catalog, longer-prefix catalog, and reasoning-auto prompt shapes were included in the expanded benchmark campaign.

## Follow-up / Out of Scope

- UI-launched parity for the exact same model/config was not part of this PR packet; proof here is API/server runtime proof.
- Cross-adapter streaming checks beyond OpenAI Chat Completions were not run for this cache fix.
- Cross-restart SSD/L2 disk-hit validation was not run.
- Mixed UI multiturn stress with VL/tools/reasoning variations was not run.
- Long soak testing was not run.
- This PR fixes cache-state correctness before broader throughput optimization work; matmul and decode-efficiency work should remain separate.

## Tests

- `pytest tests/test_mllm_scheduler_cache.py tests/test_cache_record_validator.py tests/test_native_mtp_autodetect.py -q`
- `python -m py_compile vmlx_engine/mllm_batch_generator.py vmlx_engine/mllm_scheduler.py vmlx_engine/cache_record_validator.py`
- `git diff --check`
- OpenAI Chat Completions streaming live repro/cross-quant/ablation benchmark harnesses on Qwen3.6 JANG artifacts

Full test suite was not run; validation was focused on the shared cache/native-MTP surface and live model repro matrices.


jjang-ai (commented):
Went through this against main HEAD (`70eba5877`) — diff applies cleanly, no conflicts with the qwen3_5_family vendoring or the streaming fix. Components (a)/(b)/(c) are real bugs and the fixes are directionally right; (b) in particular is well-grounded (`jang_tools/turboquant/cache.py:84` — `_idx` genuinely stays `None` until `prepare()`). But there are four blocking concerns and real test gaps before this merges:

**1. Component (a) introduces a silent long-context regression.** Routing generic hybrid-SSM stores through `_prefill_for_clean_path_dependent_cache()` replaces cheap tensor slicing (`_truncate_hybrid_cache`) with a full clean re-forward that has an ~8GB predicted-attention OOM guard (`mllm_batch_generator.py:8083-8096`, `n_heads * seq_len² * 2`). For a 32-head model that guard trips around ~11k tokens — above that, Qwen hybrid-SSM sessions silently stop getting warm paged-cache stores at all. That's a new throughput cliff this PR creates and doesn't mention, test, or measure. Please quantify it (store-path cost per completed request + the cutoff point) and either raise/handle the guard or document the cliff explicitly. Your TTFT numbers measure the warm-hit side; the store side is where this PR adds cost.

**2. Component (c) is half a fix: `_step_native_mtp_head` never got position_ids.** You updated the three backbone calls via `_native_mtp_lm_kwargs`, but `mtp_forward()` (`mllm_batch_generator.py:7247-7278`) still runs without absolute positions. If the MTP head's RoPE depends on `mtp_cache` offset the same way the backbone depended on the main cache offset, drafts on warm-hit continuations are still position-shifted — same bug class, moved into the draft head. Confirm whether qwen3_5/3.6 `mtp_forward` applies RoPE against `mtp_cache` offset, and extend the fix if so.

**3. The justification for component (d) is factually wrong, which matters for its scope.** The new comment says native MTP "needs mutable KVCache semantics" that TQ lacks — but `TurboQuantKVCache` implements `is_trimmable()`/`trim()` (`jang_tools/turboquant/cache.py:509-519`) and our rollback path already uses them. The plausible real mechanism is `trim()` decrementing `offset` without invalidating `_compressed_keys`/`_decoded_k_buffer`/`_joined_k/v` after a `compress()` boundary (step=256). If that's it, the corruption is NOT specific to warm-hit reconstruction — any cold-start native-MTP generation that crosses the compress threshold and rejects a draft is exposed, and `_native_mtp_keep_native_kv_for_prompt_cache` only guards the warm-hit entry. Please identify and cite the actual failure mode, and run a cold-start long-generation soak (JANGTQ hybrid bundle, native MTP, TQ live, no prompt-cache hit) to check whether the `!` loop reproduces without the paged cache ever being touched.

**4. The text-only scheduler path is untouched.** `native_mtp.py` aliases `qwen3_5_text`/`qwen3_6_text`, and `scheduler.py:2629-2649` delegates native MTP for text models to upstream `mlx_lm.BatchGenerator` — a completely different code path from everything patched here. Was the repro bundle VL-routed? If a text-only Qwen3.6 JANG bundle can hit the same warm-hit + TQ + MTP combination through `scheduler.py`, this PR doesn't cover it — please state which path was validated and file the gap if the text path is affected.

Smaller items:
- **Batch type mixing**: keeping one request's layers as plain `KVCache` while co-batched siblings run `TurboQuantKVCache` — does the batch merge/extend handle the type mismatch, or does that request drop out of shared-batch dispatch? Show the merge call site or confirm native-MTP requests are effectively batch-1.
- **Depth-3 positions are untested**: the new position tests pin `VMLINUX_NATIVE_MTP_DEPTH=1` or call replay directly with 2 tokens; the shipped default is depth 3, where `_run_native_mtp_verify_cycle` concatenates up to 4 tokens — the highest-risk case for an off-by-one in `_absolute_text_position_ids`. The pre-existing depth-3 test's mock cache has no `.offset`, so it silently exercises offset=0. Add a depth-3 verify-cycle test with a nonzero-offset cache asserting the exact position sequence.
- **The TQ validator test fabricates the class** (`type("TurboQuantKVCache", (), {})`) instead of importing the real one — it won't catch attribute drift in `jang_tools.turboquant`. Import the real class.
- **Please state explicitly** (and pin with a test) that the clean-prefill store re-derives through the model's TQ-patched `make_cache()`, so the keep-native exception never permanently downgrades a stored record to non-TQ form. I traced it and believe it's fine, but the PR summary invites the question and nothing pins it.

On validation: per project policy, API/server-side proof is preliminary — cache-path changes need live dev-build UI proof plus cross-adapter/SSD-restart checks. You've flagged those as out of scope yourself; they'll need to happen before this ships in a release, so let's track them on the PR rather than losing them.

PR #230, opened 2026-07-12
Guard dense Qwen native MTP from live TurboQuant KV compression

### LINKED ISSUE #229: Dense Qwen3.6 native MTP corrupts generation with live TurboQuant KV compression
## Summary

Dense Qwen3.6 generation can become severely corrupted when the MLLM path runs native MTP against a hybrid cache whose attention layers use live TurboQuant KV compression. On current `main`, a deterministic request degraded into thousands of repeated `!` characters on both a cold request and a subsequent warm `paged+ssm` hit.

The observable boundary is isolated, but the exact numerical mechanism inside MTP verification/replay and the compressed cache is not yet proven.

## Affected Scope

- Classification: engine/runtime, native MTP, and live cache compression
- Affected: dense Qwen3.6 MLLM bundles that retain native MTP tensors, use the hybrid cache path, and have live TurboQuant KV compression configured or active
- Runtime aliases: `qwen3_5`, `qwen3_5_text`, `qwen3_6`, and `qwen3_6_text`; the reproduced Qwen3.6 bundle resolves through the shared `qwen3_5` runtime family
- Gate-activation controls: the same dense bundle with live TurboQuant disabled and Qwen MoE with live TurboQuant enabled both retain active MTP; their 512-token runs reached `length` without visible content, so they are not claimed as full coherence controls
- Matrix rows in scope: A, B, K, L, M, P, U; T and V are adjacent contract checks
- Frequency: 2/2 corrupted requests on the current-main long-budget baseline; 0/2 corrupted requests with the local safety-gate candidate

### Preconditions

- vMLX `main` at `78a7aec38ab598db2fcbc7cbb6589951b813a938`
- Dense Qwen3.6 27B JANG bundle produced by a conversion that retains native MTP tensors
- MLLM hybrid-cache route
- Native MTP enabled at depth 3
- Live TurboQuant enabled with a five-token compression threshold
- Paged prefix cache enabled

### Steps

1. Start the server with native MTP and live TurboQuant enabled:

   ```bash
   VMLINUX_NATIVE_MTP=1 \
   VMLX_FORCE_TQ_AUTO=1 \
   VMLX_DISABLE_TQ_KV=0 \
   VMLX_TQ_COMPRESS_AFTER=5 \
   vmlx-engine serve <dense-qwen3.6-mtp-bundle> --native-mtp-depth 3
   ```

2. Send this OpenAI Chat Completions streaming request twice to the same server:

   ```json
   {
     "model": "<served-model-name>",
     "messages": [
       {
         "role": "user",
         "content": "give a brief summary of all the countries in south america"
       }
     ],
     "stream": true,
     "stream_options": {"include_usage": true},
     "max_tokens": 4096,
     "temperature": 0,
     "top_p": 1,
     "top_k": 0,
     "min_p": 0,
     "seed": 0,
     "repetition_penalty": 1,
     "enable_thinking": true,
     "chat_template_kwargs": {"enable_thinking": true}
   }
   ```

3. Inspect streamed reasoning, the final finish reason, native-MTP telemetry, live-TurboQuant telemetry, and cache-hit telemetry.

## Expected Behavior

Both requests should produce coherent reasoning and a complete summary. A warm cache hit must not alter generation correctness.

## Actual Behavior

Native MTP and live TurboQuant are both active. The first request reaches a repeated-`!` run of 3,587 characters; the warm request reaches 3,631. Both consume the full 4,096-token budget and finish with `length` without visible answer content. The second request is a real 15-token `paged+ssm` hit.

## Isolation

- Isolation verdict: confirmed
- Root-cause confidence: NOT PROVEN
- Failing boundary: dense Qwen shared-family MLLM route + hybrid cache + native MTP + configured or active live TurboQuant attention-KV compression
- Gate-activation controls: disabling live TurboQuant preserves native MTP; Qwen MoE preserves native MTP while live TurboQuant remains enabled. These controls prove policy exclusion and absence of a repeated-bang run within 512 tokens, not complete answer coherence.
- Separate issues excluded: answer-pass behavior at short token budgets, global TurboQuant trimming, prompt-cache record reconstruction, and a general cross-family MTP/TurboQuant claim

The failure occurs on both the cold request and the warm request, so cache loading is not required to trigger it. The evidence supports a safety boundary around the interaction as a whole, not a warm-cache-only repair.

## Root Cause / Current Finding

The exact tensor/state divergence is not proven. Current evidence shows that native MTP verification/replay is unsafe for the reproduced dense-Qwen hybrid route once live TurboQuant compression is configured or active. It does not justify rewriting TurboQuant trimming globally or changing cache reconstruction for other families.

## Proposed Fix Boundary

At the MLLM native-MTP per-request eligibility check, skip native MTP only when all of these are true:

- the runtime is hybrid;
- the exact dense-Qwen shared-family alias is selected;
- the request cache contains the canonical `TurboQuantKVCache` class from the `jang_tools.turboquant` module protocol, with compression active or scheduled.

Keep autoregressive generation, live TurboQuant, paged cache, SSM companion state, and stored q4 cache behavior unchanged. Do not match Qwen MoE, Gemma, Hy3, or generic hybrid families. Do not modify TurboQuant cache internals.

The safety policy is intentionally independent of effective MTP depth and request modality once this exact feature intersection exists. D1 was live-tested and completed coherently through the fallback; tool-capped D1 and vision requests are not live-proven to corrupt. Allowing an untested MTP submode to enter the same compressed-cache interaction would defeat the safety boundary, so unit tests pin the conservative policy before any native-MTP seed forward.

## Acceptance Criteria

- The exact cold and warm long-budget requests remain coherent while live TurboQuant stays active.
- Native MTP is skipped with an explicit reason only at the affected intersection.
- Dense Qwen with live TurboQuant disabled still activates native MTP.
- Qwen MoE with live TurboQuant enabled still activates native MTP.
- D1 and tool-capped D1 requests at the exact unsafe feature intersection are also gated before any main or MTP forward.
- A delayed live-TurboQuant threshold is gated before compression starts.
- Focused, adjacent, and full-suite checks introduce no new failing test nodes relative to the exact baseline commit.

## Evidence And Test Surface Accounting

| Surface | Result | Status | Evidence Summary |
|---|---|---|---|
| Reproduction on baseline | Both deterministic 4,096-token requests corrupted; maximum repeated-`!` runs 3,587 and 3,631 | verified | Current-main OpenAI Chat streaming run with native MTP depth 3, live q4 TurboQuant, cold miss, and 15-token warm `paged+ssm` hit |
| Candidate live validation | Both requests coherent and finished with `stop`; visible answers 2,997/2,869 chars; repeated-`!` maxima 0 and 1 | verified | Exact committed candidate on the same current-main base; native MTP skipped while live TurboQuant remained active |
| Gate-activation controls | Native MTP remained active for dense Qwen with live TurboQuant disabled and for Qwen MoE with live TurboQuant enabled | verified for policy exclusion | MTP telemetry recorded 272 and 292 cycles respectively; both 512-token runs ended at `length`, so full coherence is not claimed |
| Explicit autoregressive control | Untouched `main` with MTP explicitly disabled and live TQ unchanged produced two coherent `stop` responses | verified | Confirms the safety candidate follows the normal AR fallback behavior rather than a separate cache workaround |
| Focused regression | Candidate 105 passed versus baseline 84 in the hydrated runtime environment | verified | `pytest tests/test_native_mtp_autodetect.py -q` |
| Adjacent/shared-path checks | Candidate 293 passed versus baseline 272 with the same two unrelated failures | verified | MLLM, cache, and scheduler test selection compared on the exact same base |
| Cache/runtime proof | Cold request, warm L1 `paged+ssm` hit, live TurboQuant telemetry, native-MTP telemetry, `/health`, and memory snapshots captured | verified | Exact requests and server telemetry preserved |
| Full suite | Candidate adds 21 passing tests and has the same 46 failing node IDs as untouched `main` | verified | Candidate: 5,818 passed; baseline: 5,797 passed; identical failure-set SHA-256 |
| Live D1 | Exact long-budget prompt with native MTP configured at depth 1 completed with a coherent 3,040-character answer and `stop` | verified | Native MTP was skipped before decoding while all 16 live TurboQuant attention layers remained active |
| SSD/L2 process restart | A new process recovered 789 cached tokens through `paged+ssm+disk` and completed coherently | verified | 13 block-disk hits, one SSM-disk hit, successful reconstruction, live TQ retained, and native MTP skipped |
| Cross-adapter streaming | OpenAI Chat, OpenAI Responses, Ollama Chat, and Anthropic Messages streamed reasoning and tool-call structures | verified | Existing cross-adapter harness completed without adapter errors |
| UI-launched runtime | Candidate Electron app spawned the candidate engine and completed the exact prompt with a coherent 12-country answer | verified | 2,046 completion tokens; candidate spawn args, final health, live TQ telemetry, and exact safety log captured |
| Panel static check | TypeScript typecheck completed successfully | verified | `npm run typecheck` under the repository's expected Node 20 runtime |

## Candidate Fix Validation

A committed two-file candidate implements the proposed gate and its regression tests. On current `main` base `78a7aec3`, the exact baseline changed from two corrupted length-limited outputs to two coherent stopped outputs while preserving live TurboQuant and the warm `paged+ssm` hit. Final D1, SSD/L2 restart, cross-adapter, and UI-launched runs also passed. The candidate does not change cache trim, reconstruction, adapter, server, or UI code.

## Remaining Gaps / Out Of Scope

- The internal numerical mechanism remains unproven; this issue tracks the safety gate, not a compatibility implementation that keeps both features enabled.
- A separately converted Qwen3.5 dense artifact has not been live-tested; the reproduced Qwen3.6 runtime uses the shared `qwen3_5` alias.
- A separate text-only scheduler reproduction, live vision behavior, tool-capped D1 behavior, and a long RAM soak remain unproven or deferred and must not be described as verified.
- Disabling MTP at this intersection intentionally gives up MTP acceleration. An explicit MTP-off control was coherent, but its response lengths and timings varied relative to the candidate, so a precise speed cost is not claimed here.
- The short-budget answer-pass symptom investigated earlier is superseded by newer `main` behavior and is excluded from this issue.

jakimoai:
## Related Issue

Fixes #229

## Summary

Fall back to autoregressive decoding when native MTP intersects configured or active live TurboQuant KV compression on the dense-Qwen hybrid MLLM route. The exact current-main reproducer changed from two corrupted 4,096-token outputs to two coherent stopped outputs while preserving live TurboQuant and warm `paged+ssm` cache reuse.

## Problem

A dense Qwen3.6 27B JANG bundle that retains native MTP tensors can corrupt deterministic generation when all of these are active:

- the MLLM hybrid-cache route;
- native MTP at depth 3;
- live TurboQuant on attention KV layers.

On current `main` at `78a7aec38ab598db2fcbc7cbb6589951b813a938`, the exact OpenAI Chat streaming request failed on both the first cold request and the repeated warm request. The reasoning streams ended with repeated-`!` runs of 3,587 and 3,631 characters, exhausted the 4,096-token budget, and produced no visible answer. The second request was a real 15-token `paged+ssm` hit.

## Root Cause

The observable interaction boundary is proven; the exact numerical mechanism is **NOT PROVEN**. Native MTP verification/replay is unsafe on the reproduced dense-Qwen hybrid route once live TurboQuant compression is configured or active. The evidence does not justify a global TurboQuant trim change or a cache-reconstruction rewrite.

## Fix

- Add an exact dense-Qwen runtime alias set for the shared Qwen3.5/3.6 MLLM implementation.
- Detect only canonical `TurboQuantKVCache` layers from the `jang_tools.turboquant` module protocol with already-compressed tokens or a positive future compression threshold.
- Pass the actual request cache into the existing per-request native-MTP eligibility check.
- Skip native MTP with an explicit reason only when dense Qwen, hybrid cache, and configured/active live TurboQuant all intersect.
- Add focused inclusion, exclusion, delayed-threshold, gate-precedence, pre-forward, and real TurboQuant-class regressions.

## Why This Shape

This PR is intentionally a safety fallback, not a speculative cache rewrite. It leaves live TurboQuant, paged prefix caching, SSM companion state, stored q4 cache behavior, and normal autoregressive generation in place.

It does not match Qwen MoE, Gemma4, Hy3, non-hybrid runtimes, dense Qwen with live TurboQuant disabled, or TurboQuant object-only mode. It does not modify cache trim, compression, reconstruction, adapters, server endpoints, or UI code.

Inside the exact unsafe feature intersection, the fallback deliberately applies before effective MTP depth or modality-specific generation. D1, tool-capped D1, and vision corruption are not live-proven; they remain conservatively gated until compatibility is established rather than being allowed into the same unproven interaction.

## Live Proof

| Interface / Harness | Model / Config | Result |
|---|---|---|
| OpenAI Chat Completions streaming, cold request | Dense Qwen3.6 27B JANG_4M, MLLM hybrid cache, MTP D3, live TQ threshold 5 | Baseline: 3,587 repeated bangs and `length`; candidate: coherent 2,997-character answer and `stop` |
| Same request, warm cache | Same server; 15-token `paged+ssm` hit | Baseline: 3,631 repeated bangs and `length`; candidate: coherent 2,869-character answer and `stop` |
| Live D1 safety route | Same dense artifact and exact prompt; native MTP configured D1; live TQ retained | Candidate skipped MTP before decoding and returned a coherent 3,040-character answer with `stop` |
| SSD/L2 process restart | 790-token prompt primed, server stopped, new candidate process replayed the same prompt | 789-token `paged+ssm+disk` hit; 13 block-disk hits plus one SSM-disk hit; coherent `stop` response |
| Cross-adapter streaming | OpenAI Chat, OpenAI Responses, Ollama Chat, and Anthropic Messages | All four streamed reasoning and tool-call structures through the gated MLLM generator without adapter errors |
| UI-launched session | Candidate Electron dev app; detected D3 native MTP and live TQ; exact prompt through app chat | Coherent 12-country answer; 2,046 completion tokens; final health retained live TQ and recorded no native-MTP cycle |
| Dense-Qwen activation control | Same dense artifact, live TQ disabled, stored q4/paged cache retained | Native MTP remained active; final request recorded 272 cycles; 512-token output ended at `length`, so full coherence is not claimed |
| Model-family activation control | Qwen3.5 MoE 35B-A3B JANG_1L, live TQ enabled | Native MTP remained active; final request recorded 292 cycles and 0.693 acceptance; 512-token output ended at `length` |
| Explicit AR control | Dense Qwen3.6 27B JANG_4M, MTP explicitly off, live TQ retained | Two coherent `stop` responses, including the same 15-token warm hit |

Evidence captured:

- spawn args: yes
- startup logs: yes
- `/health`: yes
- streaming reasoning/content and finish reasons: yes
- native-MTP and live-TurboQuant telemetry: yes
- cold and warm L1 cache state: yes
- SSD/L2 restart with block and SSM disk hits: yes
- UI-launched candidate session: yes
- OpenAI Responses, Ollama, and Anthropic streaming: yes
- memory snapshots: yes
- RAM soak: no; deferred below

## Regression / Adjacent Coverage

- Dense Qwen with live TQ disabled proves this is not a blanket dense-Qwen MTP disable.
- Qwen MoE with live TQ enabled proves the family exclusion on a real MTP/TQ runtime.
- Direct canonical-class tests pin configured, inert, and active states; a same-name non-JANG decoy is excluded.
- Delayed compression is gated before MTP begins, preventing a request from entering MTP and crossing the TQ threshold later.
- D1 and tool-capped D1 unit paths are conservatively gated before any main/MTP seed forward; live D1 also completed coherently.
- Candidate and untouched baseline have identical focused-adjacent and full-suite failure nodes.

## Follow-up / Out of Scope

- Determine the precise MTP verify/replay versus compressed-cache state divergence and implement a true compatibility fix that can retain both features.
- Live-test a separate dense Qwen3.5 MTP artifact; the reproduced Qwen3.6 route resolves through the shared `qwen3_5` runtime alias.
- Live-test vision behavior, a separate text-only scheduler reproduction, and a longer RAM soak before making those broader claims.
- Quantify the precise performance cost of autoregressive fallback. The explicit MTP-off control is coherent, but differing completion lengths and run-to-run timings do not support a percentage claim.

## Tests

- `pytest tests/test_native_mtp_autodetect.py -q` (candidate `105 passed`; untouched baseline `84 passed` in the hydrated runtime profile)
- `pytest tests/test_native_mtp_autodetect.py tests/test_hybrid_live_tq_kv.py tests/test_mllm.py tests/test_mllm_continuous_batching.py tests/test_mllm_scheduler_cache.py -q` (candidate `293 passed`; untouched baseline `272 passed`; the same two existing failures)
- `pytest -q` (candidate `5,818 passed`; untouched baseline `5,797 passed`; the same 46 failing node IDs, 102 skips, and 92 deselections)
- `python -m py_compile vmlx_engine/mllm_batch_generator.py tests/test_native_mtp_autodetect.py`
- `git diff --check`
- `npm run typecheck` in the candidate panel (`pass`)
- `cross_adapter_stream.py` against the live D1 candidate (`pass` across OpenAI Chat, Responses, Ollama, and Anthropic)
- SSD/L2 stop/restart replay (`paged+ssm+disk`, 789 cached tokens, coherent `stop`)
- UI-launched exact-prompt run from the candidate worktree (`pass`, coherent 12-country answer)
- Targeted Ruff candidate/baseline comparison: both report the same 351 existing findings; this diff adds no finding
- Exact live OpenAI Chat streaming before/after and gate-activation controls described above

Raw focused, adjacent, full, failure-list, static, Ruff, environment, and commit-patch artifacts were preserved with hashes before packet finalization.


jjang-ai:
Holding this one — please see my analysis in #229 first. The short version: I drove the real `TurboQuantKVCache` through the exact MTP verify→trim→replay shape and **TQ's rewind is content-correct** (bit-identical to `KVCache` except the single encoded token, with or without MTP). So the intersection this PR fences off is not corrupting the cache the way the issue assumes, and the fix is scoped to a mechanism nobody has identified yet.

Gate-specific review, since the placement work here is actually good and I want you to keep it:

**Correct:** the call site in `_seed_native_mtp_from_prefill` (`mllm_batch_generator.py:7415`) returns before the seed forward (`:7429`), before `make_mtp_cache()` (`:7443`) and before `_native_mtp_depth_for_request()` (`:7444`). Depth-1 and tool-capped depth-1 are genuinely covered, and it is depth-independent as claimed. That part does what it says.

**The family list is arbitrary and partly dead.**
- `qwen3_5_moe` / `qwen3_5_moe_text` are **on the TQ allow-list** (`utils/hybrid_tq_cache.py:9-16`, gate `:51-64`) — MoE reaches the *identical* intersection. So do `nemotron_h`/`nemotron_h_v2` (`:18-23`) and gemma-4 mixed-SWA (`is_mixed_swa_tq_supported`, `:84-96` → `jang_loader.py:1800-1812`). They're excluded on the strength of a control the issue itself concedes isn't a coherence control (512 tokens, no visible content).
- `qwen3_6` / `qwen3_6_text` are **dead entries** — they aren't in `QWEN36_HYBRID_MODEL_TYPES`, so a config reporting those types never gets hybrid TQ installed at all. The alias set was never diffed against `hybrid_tq_cache.py`.
- The gate requires `self._is_hybrid` (`:3965-3970`), so **dense qwen3.5 on the plain non-hybrid TQ path** (`jang_loader.py:1826-1832`) — same TQ cache, same MTP, no SSM — is ungated.

**Three fail-open holes:**
1. `cache: list[Any] | None = None` with `if cache is not None and ...` — the safety check is opt-in at the signature level; any caller that omits `cache` silently bypasses it.
2. `_is_jang_turboquant_cache` requires `type(layer).__module__.startswith("jang_tools.turboquant.")`. A subclass, a vendored re-export, or the **bundled-python** copy under a different package path → detection fails → **MTP proceeds**. This contradicts the comment you leave in place three lines up (`_TQ_CLASS_NAME` … *"enables KV-like detection without importing TQ"*, `:1429-1433`) and diverges from every other TQ detector here, which are name-only (`_is_kv_like` `:1445`; `turboquant_config.py:167`). `test_same_name_non_jang_cache_is_excluded` **pins this fail-open as intended**.
3. `_cache_has_configured_or_active_turboquant` iterates the cache list **flat** — it never descends into `.caches` children, unlike `turboquant_cache_telemetry.visit()` (`turboquant_config.py:166-173`) and `_ensure_batch_cache`'s `_CacheList` handling (`:3573`). A nested TQ layer → gate off.

**A second MTP engine is left completely ungated.** There are two native-MTP implementations; this gates only the MLLM one. The text/oMLX path (`patches/mlx_lm_mtp/batch_generator.py`, activated via `scheduler.py:2733-2735`) runs against TQ caches too — `scheduler.py:382-386` tracks `_hybrid_live_tq_compress_after` for exactly that path. A `qwen3_5_text` bundle served through the LLM route hits the same intersection and is untouched by this PR. Given your own rationale ("independent of effective MTP depth and request modality once this exact feature intersection exists"), leaving the second engine open is internally inconsistent. If a policy like this ever lands, its home is `native_mtp.py` — which already owns `native_mtp_blocked` and governs both paths.

**The tests pin the symptom, not the mechanism.** `test_request_gate_disables_only_the_proven_intersection` and `test_non_dense_qwen_aliases_are_excluded` *assert* that `qwen3_5_moe`, `gemma4` and `hy_v3` keep native MTP with an active TQ cache. If the mechanism is generic — and you've said it's unknown — those tests actively block the correct fix later.

**It also conflicts with your own #223**, which claims a *proven* mechanism for the same symptom on the same intersection (warm reconstructed prompt-cache re-compressed in `_rebuild_turboquant_from_reconstructed`, `mllm_batch_generator.py:1723-1775`) and repairs the representation rather than disabling the feature. Both edit the same `_native_mtp_*` region of `mllm_batch_generator.py` and the same test file, and #230 would make #223's warm-cache scenario unreachable on the family it was proven against. These are two fixes for one defect; they can't both land.

**What I'd rather you do:** run the missing control (dense Qwen + live TQ + native MTP **off**), then chase `_update_and_fetch_batched` (`jang_tools/turboquant/cache.py:227-251`) — it reads `self.keys` which `compress()` sets to `None`, allocating a fresh zero buffer that drops the whole prefix → attention over ~zero KV → token id 0 → `"!"`. That is the symptom you reported, and it's the only path in TQ I found that produces it.

The evidence discipline in the writeup is genuinely strong — I just don't want to fence off a family name in place of a mechanism.

PR #248, opened 2026-08-02
DSV4: fail-closed affine decode integration and native stability paths

hornsan1:
## Summary

Harden the DeepSeek-V4 affine decode integration and add guarded native-MLX stability paths with reproducible exact-output gates.

The original timing-only submission has been extended to address the review blockers:

- Loader-owned DSV4 installation with atomic layout validation.
- Single-token decode ownership with prefill and unsupported-shape fallback.
- Preserved `[B, S, K, H]` route rank.
- Runtime exception fallback and per-module disable behavior.
- Idempotent installation and stale-module selection clearing.
- Populated numeric parity coverage for the affine route.
- Native compiled weighted-gather MoE decode path.
- Native quantized lm-head path with a bounded FP32 opt-in fallback.
- Bounded exact RoPE table reuse; native RoPE remains opt-in.
- Request-bounded unused-indexer bypass with native fallback.
- Portable 30-token and 1,000-token harnesses with optional expected stream-hash gates.

## Why

The DSV4 JANG runtime is sensitive to cache state, long-run arithmetic drift, and mixed affine layouts. Short throughput results are not sufficient evidence: a candidate must preserve the token stream and must not turn a runtime/compiler mismatch into a production request failure.

The default path remains fail-closed. Unsupported layouts, prefill shapes, cache-hit paths, malformed state, and optimized dispatch failures retain the native implementation.

## Validation

Environment:

- Apple M4 Max, 40 GPU cores, 128 GB unified memory
- MLX 0.32.0
- mlx-lm 0.31.3
- JANG 2.5.39
- DeepSeek-V4-Flash mixed-affine JANG bundle

Focused no-model suite:

- **87 passed, 1 skipped**
- Includes affine fast-path parity/fallback, stability fast paths, batch-generator, DSV4 contract, route-mode, and bundle-integrity tests.

Real-model target gate:

- 3 repeats, 30 generated tokens each
- Median decode: **27.12 tok/s**
- Median prefill: **231.01 tok/s**
- All streams matched reference hash `fc38a393f08f90ccf8e0daab774f33a6c2d67090b43f471eeb45f0e3c50d86c1`

Real-model long-run gate:

- 1,000 generated tokens
- Decode: **27.03 tok/s**
- Prefill: **233.58 tok/s**
- Exact reference hash: `d3df86b62ad4fc14a0a0be8f6e08261417ef4002c630d1778c10ad87e070e3b1`

The 30 tok/s objective remains open; this PR does not claim to have reached it.

## Reproduction

See `docs/development/dsv4-decode-acceptance.md`.

The important boundary is to pass `--expected-token-hash` when a run is being used as an exact-reference acceptance gate. A repeatable stream without a supplied reference is reported separately and is not promoted as exactness evidence.

## Additional check

The broader `tests/test_dsv4_*.py` family reached 265 passed and 2 skipped, with one unrelated pre-existing panel-source contract failure on the PR base branch. The focused DSV4 suite and both real-model gates pass.

### DIFF
--- a/bench/affine_moe_decode_bench.py
+++ b/bench/affine_moe_decode_bench.py
@@ -1,0 +1,125 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+"""Affine MoE decode kernel microbench.
+
+Synthetic benchmark — does NOT load a model. Exercises the fused
+dequant+matvec kernels from vmlx_engine.metal.affine_moe_decode against
+MLX's stock gather_qmm baseline, using DeepSeek-V4-Flash JANG shapes.
+
+Usage:
+    python bench/affine_moe_decode_bench.py [--experts 256] [--k 6] [--iters 200]
+
+Reports per-projection latency and total MoE-layer estimate.
+"""
+
+from __future__ import annotations
+
+import argparse
+import time
+from typing import Any, Callable
+
+import mlx.core as mx
+
+from vmlx_engine.metal.affine_moe_decode import AffineMoEDecodeManager
+
+
+def _sync(*values: Any) -> None:
+    mx.eval(*values)
+    mx.synchronize()
+
+
+def _time_ms(fn: Callable[[], Any], *, warmup: int, iters: int) -> float:
+    for _ in range(warmup):
+        _sync(fn())
+    t0 = time.perf_counter()
+    for _ in range(iters):
+        _sync(fn())
+    return (time.perf_counter() - t0) * 1000.0 / max(1, iters)
+
+
+def _make_affine_weights(
+    n_experts: int, out_dim: int, in_dim: int, bits: int, group_size: int
+) -> tuple[mx.array, mx.array, mx.array]:
+    """Create synthetic affine-quantized weights matching MLX layout."""
+    n_groups = in_dim // group_size
+    vals_per_u32 = 32 // bits
+    packed_cols = in_dim // vals_per_u32
+    code_mask = (1 << bits) - 1
+
+    weights = mx.random.uniform(
+        shape=(n_experts, out_dim, packed_cols), low=0, high=2**32 - 1
+    ).astype(mx.uint32)
+    # Mask to valid code range
+    weights = weights & code_mask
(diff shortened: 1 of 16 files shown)


jjang-ai:
Thanks for putting together the DSV4 affine decode fast path and including the measured target. I checked this against the current Python runtime and the exact DSV4 affine layout. There are several blockers before this can safely land:

1. `install_affine_moe_fastpath()` has no production caller. As submitted, merging the PR cannot change runtime throughput.
2. If a caller is added, the install replaces `SwitchGLU.__call__` globally. The decode branch assumes affine `scales`, `biases`, `bits`, and `group_size`, and unsupported patterns/Metal failures raise instead of falling back to stock. This needs DSV4-affine instance/loader scoping, validation, idempotence, and a real exception fallback.
3. The replacement drops the sequence dimension: stock `x=[B,S,H], indices=[B,S,K]` returns `[B,S,K,H]`, while the `x.ndim == 3` branch reshapes to `[B,K,H]`.
4. The benchmark is timing-only and its packed-weight fixture masks each whole uint32 by one code mask, leaving nearly all packed codes zero. Please add populated packed-weight numeric parity tests against `mx.gather_qmm` for b2/g64, b3/g64, and b2/g32, plus unsupported-shape/fallback, prefill, and repeated-install cases.
5. `groups_per_chunk = n_groups / n_chunks` silently drops remainder groups. The current DSV4 keeper happens to have 64 groups, but the kernel must guard divisibility and fall back for other shapes.

The exact DSV4 keeper layouts do match the three intended kernels, so the approach remains worth pursuing. The minimum safe path is a loader-owned DSV4-affine install, exact module/shape/mode guards, preserved rank, true fallback/idempotence, populated numeric-parity tests, and then an exact-bundle no-cache decode A/B. This PR is independent of prefix-cache/reminder serialization and should not be used as evidence for those fixes.

hornsan1:
Superseded by #249. The original branch was not promoted: its direct quantized-head path diverged at the 1,000-token hash gate, its benchmark surface was not source-pinned, and it mixed unrelated runtime-default experiments. #249 rebuilds the useful work on current `main` with exact arithmetic, bounded instance-scoped caches, allocator-budget accounting, fail-safe fallback, and sealed native-versus-optimized evidence.

PR #249, opened 2026-08-04
perf(dsv4): cache exact decode projections safely

hornsan1:
Supersedes #248 with a clean implementation on current `main`.

## What changed

- Cache the exact dequantized FP32 DSV4 vocabulary head once, then reuse the same FP32 matmul as JANG.
- Gate the cache on model geometry, effective Metal headroom, and a 4 GiB allocator safety floor.
- Deduct the persistent head bytes from the existing MLX allocator-cache ceiling. On the tested model, the 2,118,123,520-byte head lowers the ceiling from 24,739,011,624 to 22,620,888,104 bytes rather than expanding the runtime memory envelope.
- Scope dispatch to weakly registered model instances. Reconfiguration, unload, env disablement, or validation failure releases the reservation.
- Fall back from the already-computed hidden state, so a failure never replays the transformer or advances a mutable request cache twice.
- Share the existing exact manual RoPE tables only across validated equal-frequency instances, with bounded LRU tables and weak cleanup.
- Add a checkout-pinned native-versus-optimized acceptance suite and sealed, public-safe raw reports.

This intentionally does **not** promote compiled MoE, unused-indexer mutation, native RoPE, the 512-token layerwise-prefill threshold, or the old custom affine microbenchmark. The guarded affine Metal path already on `main` remains separately controlled.

## Rejected during iteration

- Direct quantized-head dispatch was faster but changed the 1,000-token stream hash.
- A cached FP16 head preserved hashes but produced severe residency stalls.
- An FP32 cache without allocator reservation was exact but intermittently collapsed to ~0.8-1.2 tok/s at 400/512-token prompts.
- Early benchmark runs could import the installed vMLX package instead of the checkout. The harness now pins and verifies its source root.
- Native runs can also show allocator cliffs. Acceptance rejects both repeat-level tails and consistently pathological context bands; it cannot use an unstable native run to inflate candidate ratios.
- A synthetic allocator-cold protocol was tested and rejected because it made short decode consistently pathological. The accepted matrix is warm and sequential, matching the runtime environment.

## Sealed result

Implementation commit: `15e9420db804cfa81167acaceae13476d856a7b5`.

Bundle identity: `OsaurusAI__DeepSeek-V4-Flash-0731-JANG`, 102 shards / 101,982,500,132 bytes, manifest `fc68310e3b81ccbf5b86b9c28929b629dd89fa204d5de563d5442b13f6f2978c`.

Hardware/runtime: arm64 macOS 26.5.1, MLX 0.32.0, mlx-lm 0.31.3, JANG 2.5.39, vMLX 1.6.22.

| Case | Decode ratio | Slowest-sample ratio | Prefill ratio | Exact hash |
|---|---:|---:|---:|:---:|
| p256 / 30 | 1.293x | 1.227x | 0.983x | yes |
| p400 / 30 | 1.241x | 1.210x | 0.996x | yes |
| p512 / 30 | 1.195x | 1.185x | 0.997x | yes |
| p1024 / 30 | 1.186x | 1.184x | 1.021x | yes |
| p2048 / 30 | 1.315x | 1.325x | 1.008x | yes |
| p256 / 1000 | 1.273x | 1.273x | 1.057x | yes |

Raw reports:

- `reports/pr249/dsv4-native-15e9420d.json`
- `reports/pr249/dsv4-optimized-15e9420d.json`

Both reports identify implementation commit `15e9420d`. Baseline compatibility, feature-state proof, bundle/source identity, exact hashes, repeat-tail checks, cross-context sanity, and performance thresholds all pass. Reports contain no machine-local paths.

## Public repository hygiene

The branch was rewritten before republishing so the forbidden development-doc path and private report paths are absent from its public history. The CI baseline repair:

- exact-allowlists only the two DSV4 contract documents already present on public refs;
- sanitizes six pre-existing machine-specific path literals instead of exempting private paths;
- adds a regression test for the exact allowlist;
- passes `scripts/check-public-repo-hygiene.sh` locally against all advertised public refs.

## Validation

- `301 passed, 2 skipped` across all DSV4 unit/regression tests, the answer-pass family, and public-hygiene tests
- Ruff: clean on all new/modified Python candidate, harness, probe, and test files
- Compileall: clean
- Full public-repository hygiene script: pass
- `git diff --check origin/main...HEAD`: clean
- Private-path scan over the PR diff: clean
- Panel Vitest was unavailable locally because `panel/node_modules` is not installed; the affected test only replaces a hard-coded private missing path with an equivalent `os.tmpdir()` path.

### DIFF
--- a/bench/dsv4_api_tool_cache_proof.py
+++ b/bench/dsv4_api_tool_cache_proof.py
@@ -101,7 +101,7 @@ def main():
     base_msgs = [
         {"role": "system", "content": "You are a helpful coding agent."},
         {"role": "user",
-         "content": "Read the file /Users/eric/r21-dsv4-work/r21_probe_one.txt "
+         "content": "Read the file /workspace/r21_probe_one.txt "
                     "and tell me exactly what it contains."},
     ]
 
--- a/bench/dsv4_fallback_probe.py
+++ b/bench/dsv4_fallback_probe.py
@@ -8,11 +8,15 @@ documented cause of the unbounded literal `response` loop.
 """
 import os
 import sys
+from pathlib import Path
 
-sys.path.insert(0, os.environ.get("VMLX_SRC", "/Users/eric/mlx/vllm-mlx-r20-dsv4"))
+sys.path.insert(
+    0,
+    os.environ.get("VMLX_SRC", str(Path(__file__).resolve().parents[1])),
+)
 
 from vmlx_engine.api.tool_calling import check_and_inject_fallback_tools  # noqa: E402
-from vmlx_engine.loaders import dsv4_chat_encoder as E  # noqa: E402
+from vmlx_engine.loaders import dsv4_chat_encoder as E  # noqa: E402, N812
 
 sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
 from dsv4_prefix_probe import MODEL_PATH, TOOLS, TURNS  # noqa: E402
--- a/bench/dsv4_interleaved_probe.py
+++ b/bench/dsv4_interleaved_probe.py
@@ -27,13 +27,17 @@ import copy
 import json
 import os
 import sys
+from pathlib import Path
 
-sys.path.insert(0, os.environ.get("VMLX_SRC", "/Users/eric/mlx/vllm-mlx-r20-dsv4"))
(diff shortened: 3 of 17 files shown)


hornsan1:
Outdated

PR #265, opened 2026-08-31
Support current mlx-vlm video loading

hornsan1:
## Summary
- preserve the legacy mlx-vlm 0.5 `fetch_video` path
- fall back to the current `mlx_vlm.utils.load_video` API
- cover both a missing legacy module and a removed legacy symbol

## Why
Current mlx-vlm releases removed `video_generate.fetch_video`, causing video requests to fail before processor ingestion.

## Tests
- `.venv/bin/python -m pytest -q tests/test_qwen4_exp_runtime.py -k "current_mlx_vlm_video_loader or qwen_video_float_255"` (3 passed)

### DIFF
--- a/tests/test_qwen4_exp_runtime.py
+++ b/tests/test_qwen4_exp_runtime.py
@@ -1497,6 +1497,41 @@ def test_qwen_video_float_255_input_preserves_uint8_processor_scale():
     assert untouched is unit_interval
 
 
+@pytest.mark.parametrize("legacy_module", [None, object()])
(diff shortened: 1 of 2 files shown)


jjang-ai:
Verified against current main: the compat helper applies cleanly, the exception scope is correctly limited to import/symbol resolution so real decoder errors still propagate, and the full qwen4_exp runtime suite passes with the two new loader tests. Merged. Thanks for the clean fix.

PR #266, opened 2026-08-31
Preserve JANG sparse cache during batching

hornsan1:
## Summary
- add a batch-aware MiniMax-M3/JANG sparse cache that preserves K/V state and `idx_keys`
- use it for multi-request cache merging and single-request batch promotion
- keep sparse index history aligned across merge, extract, filter, extend, trim, and finalize operations

## Why
Generic batching converted `MiniMaxM3SparseCache` into `BatchKVCache`, discarding `idx_keys`. Concurrent Qwen/JANG decode then failed because the promoted cache no longer provided `update_index`.

## Tests
- `.venv/bin/python -m pytest -q tests/test_minimax_m3_batch_cache.py tests/test_minimax_m3_cache_paths.py tests/test_mllm_continuous_batching.py` (101 passed, 3 deselected)

### DIFF
--- a/tests/test_minimax_m3_batch_cache.py
+++ b/tests/test_minimax_m3_batch_cache.py
@@ -1,0 +1,137 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Regression coverage for Qwen/JANG sparse-cache continuous batching."""
+
+import pytest
+
+mx = pytest.importorskip("mlx.core")
+
+from vmlx_engine.mllm_batch_generator import (  # noqa: E402
+    _ensure_batch_cache,
+    _merge_caches,
+)
+from vmlx_engine.models.minimax_m3.cache import (  # noqa: E402
+    BatchMiniMaxM3SparseCache,
+    MiniMaxM3SparseCache,
+)
+
+
+def _fill_sparse(length: int, base: float) -> MiniMaxM3SparseCache:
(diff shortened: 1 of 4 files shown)


jjang-ai:
Reviewed and verified against current main before merging: the patch applies cleanly at head; behavioral A/B confirms the pre-fix defect (generic promotion turns MiniMaxM3SparseCache into a plain BatchKVCache with no idx_keys and no update_index, so the first concurrent decode step fails) and the specialized batch cache preserves the indexer lane through merge, extract, filter (including the left-padding shift), extend, trim, and finalize, with the lockstep and batch-shape invariants failing closed. All 5 new tests pass plus the M3 cache/loader/parser suites (48) and the shared mllm batching suites with no new failures. Merged. Live two-request batching proof on a real M3 bundle is queued on our side. Thanks for the thorough fix and tests.

jjang-ai:
Live-proof follow-up: on current main the engine scheduler force-overrides MiniMax-M3 to the single-sequence native path even when launched with --max-num-seqs 2 (minimax_m3_vl path-dependent state policy), so the generic batch promotion this PR guards cannot currently be reached at runtime; two concurrent requests serialize and complete coherently. The fix stands as the correct fail-closed seam for whenever that override is lifted, and the unit matrix fully covers the lanes. If your setup batched M3 on an earlier build, that path is what this protects.

